PR #5, opened 2016-07-22
Set the disk number to 0xFFFF when Zip64 extra field is involved

julik:
to make things work with The Unarchiver 3.10

This also rejigs the microzip test for more verbosity (see commit messages for an explanation)


dsnipe:
👍

PR #6, opened 2016-10-10
Clean and uniqify streamed filenames

dsnipe:


lib/zip_tricks/streamer.rb
```diff
@@ -241,28 +239,48 @@ def close
...
   def add_file_and_write_local_header(filename:, crc32:, storage_mode:, compressed_size:,
       uncompressed_size:, use_data_descriptor: false)
-    if @files.any?{|e| e.filename == filename }
-      raise DuplicateFilenames, "Filename #{filename.inspect} already used in the archive"
-    end
-    
+
+    # Clean backslashes and uniqify filenames if there are duplicates
+    filename = remove_backslash(filename)
+    filename = uniqify_name(filename) if @files.any? { |e| e.filename == filename }
```
julik: Since it is a recursive algorithm, I think it might be better to just do it regardless, without the conditional

lib/zip_tricks/streamer.rb
```diff
@@ -241,28 +239,48 @@ def close
...
   end
+
+  def remove_backslash(filename)
+    filename.tr('\\', '_')
+  end
+
+  def uniqify_name(filename)
+    copy_pattern = /\((\d+)\)$/ # we add (1), (2), (n) at the end of a filename if there is a duplicate
+    parts = filename.split(".")
+    ext = parts.pop if parts.size > 1
```
julik: Do you account for files with double extensions here, like `.txt.gz`?

lib/zip_tricks/streamer.rb
```diff
@@ -241,28 +239,48 @@ def close
...
+    filename_end = parts.pop
+    if filename_end =~ copy_pattern
```
julik: This can be turned into a `loop` with an implicit counter starting at 1, then you can do both the substitute, the test for inclusion and the counter increment in one place

lib/zip_tricks/streamer.rb
```diff
@@ -241,28 +239,48 @@ def close
...
   def write_data_descriptor_for_last_entry
     e = @files.fetch(-1)
     @writer.write_data_descriptor(io: @out, crc32: 0, compressed_size: e.compressed_size, uncompressed_size: e.uncompressed_size)
   end
+
+  def remove_backslash(filename)
+    filename.tr('\\', '_')
+  end
+
+  def uniqify_name(filename)
```
julik: It's possible to make it more functional by allowing the method to accept an Enumerable of filenames already present. Then you can also turn it into a Set, which has (presumably) cheaper include checks (which you might have to do quite a few of).

spec/zip_tricks/streamer_spec.rb
```diff
@@ -276,4 +276,14 @@ def rewind_after(*ios)
...
     }.not_to yield_control
   end
+
+  it 'prevents duplicates in the stored files' do
+    files = ["file.one\\two.jpg", "file_one.jpg", "file_one (1).jpg", "file\\one.jpg"]
+    zip_streamer = described_class.new(StringIO.new)
+    files.each do |fn|
+      zip_streamer.add_stored_entry(filename: fn, size: 1024, crc32: 0xCC)
+    end
+    zip_streamer_file = zip_streamer.instance_variable_get("@files")
```
julik: You don't have to do this. `Streamer` supports an injectable `writer:` now. You can supply it with a double and then assert on the double that it gets the right calls (`expect(w).to receive(:write_local_file_header).with(...)`

lib/zip_tricks/streamer.rb
dsnipe: It's not clear for me. If I'll remove the conditional, it'll call every time and we won't go out from the recursion.

lib/zip_tricks/streamer.rb
dsnipe: Well, I need to know all possible double extensions here if we want to solve that case. I'll check, maybe I'll find the list of them.

lib/zip_tricks/streamer.rb
julik: You can loop within the function. For example:

```
duplicate_i = 0
loop do
  ...
  return unless filenames_in_archive.include?(...)
  duplicate_i += 1
end
```

lib/zip_tricks/streamer.rb
julik: Not necessarily. Any number of `/\.\w+` at the end qualifies as an extension IMO

julik: I'd put the ` (1)` before the `.txt.zip` so that it would become `My.Favorite.Artists (1).txt.zip` - so the `\w` is prolly more like `[a-z\d]`. I agree it's murky water tho ;-)

PR #11, opened 2017-04-03
adds ability to create empty directories

bermannoah:
Adds the ability to create an empty directory during the streaming process and sets the permissions relatively sanely. Have run the tests I was able to given my machine's limits on macOS and then tested the new functionality on Windows. Let me know if you'd like a different default set of permissions for the empty directory. Looking forward to discussing this with you!


lib/zip_tricks/streamer.rb
```diff
@@ -202,7 +202,13 @@ def write_deflated_file(filename)
     last_entry.uncompressed_size = uncomp
     write_data_descriptor_for_last_entry
   end
-
+  
+  # Adds an empty directory to the archive.
+  def add_empty_directory(filename:, size:0, crc32:0)
```
julik: The `size:` and the CRC will be ignored for directories since they have no "body" content. These arguments can just be removed and set hardcoded in the subsequent `add_file_and_write_local_header` call

lib/zip_tricks/zip_writer.rb
```diff
@@ -176,7 +182,14 @@ def write_central_directory_file_header(io:, local_file_header_location:, gp_fla
       io << [0].pack(C_v)
     end
     io << [0].pack(C_v)                                # internal file attributes        2 bytes
-    io << [DEFAULT_EXTERNAL_ATTRS].pack(C_V)           # external file attributes        4 bytes
+    
+    # Because the add_empty_directory method will create a directory with a trailing "/",
+    # this check can be used to assign proper permissions to the created directory.
+    if filename[-1] == "/"
```
julik: Since we only support modern Rubies we could also use `String#end_with?`

lib/zip_tricks/streamer.rb
julik: Also, I would personally change the `filename:` kwarg to `dirname:` to make it more obvious that this is a directory name, but that's more of a taste thing.

spec/zip_tricks/streamer_spec.rb
```diff
@@ -142,6 +142,50 @@ def rewind_after(*ios)
...
+
+      outbuf.flush
+      File.unlink('test.zip') rescue nil
+      File.rename(outbuf.path, 'osx-empty-test.zip')
+
+      # Mark this test as skipped if the system does not have the binary
+      open_zip_with_archive_utility('osx-empty-test.zip', skip_if_missing: true)
+    end
+    FileUtils.rm_rf('osx-empty-test')
+    FileUtils.rm_rf('osx-empty-test.zip')
```
julik: If the spec fails to run, this will leave a file on the filesystem. There is kind-of-a workaround for this in the other test that opens the file with ArchiveUtility, and you can add the call for the empty directory into that test

testing/generate_test_files.rb
```diff
@@ -8,6 +8,16 @@
   zip << $image_file
 end
 
+build_test "Two small stored files and an empty directory" do |zip|
```
julik: Adding it to the big test is cool, really like that

julik:
Awesome stuff! To drive it home - see a few comments inline. Additionally:

* The same method could be useful in the `SizeEstimator` (these headers also take up space)
* This adds an implicit switch in behavior within the Writer class, so a test that actually unpacks the external attributes could do very well in the `writer_spec...`
* Despite what it says on the tin, the `write_central_directory_file_header` has a bug (or rather an omission). It accepts an `external_attrs` argument, but that argument is actually ignored. It's set to `123` in multiple Writer tests, but if you actually use it the tests will fail. I think given what has been done we should remove that argument completely since we only support 2 hardcoded values, and reintroduce it only if it turns out someone needs (and uses) that argument
* `add_empty_directory` should be documented with YARD like the rest of the methods
* IMO it should also return the IO offset like the other methods do (for consistency)
* Maybe a smaller-scoped test for the directory write only could be added to the Streamer spec (checking for one entry) or to an existing example

lib/zip_tricks/streamer.rb
bermannoah: Just pushed up these changes!

lib/zip_tricks/zip_writer.rb
bermannoah: And this change!

spec/zip_tricks/streamer_spec.rb
bermannoah: I've moved the empty directory call in the ArchiveUtility test but it still seems to be opening and leaving a file in my Downloads directory. I've tried hardcoding in the path to that directory but it doesn't seem to want to delete the extra file.

bermannoah:
@julik -- thanks again for the feedback! I've pushed up the changes -- added the method to the `SizeEstimator` and the suggested tests, then went through and cleaned up the tests that were changed by the `external_attrs` removal, etc. I think I added the YARD docs correctly but let me know if it (or anything else) should change! Looking forward to talking with you!

lib/zip_tricks/size_estimator.rb
```diff
@@ -61,4 +61,13 @@ def add_compressed_entry(filename:, uncompressed_size:, compressed_size:, use_da
     @streamer.write_data_descriptor_for_last_entry if udd
     self
   end
+
+  def add_empty_directory_entry(dirname:, use_data_descriptor: false)
```
julik: For empty directories data descriptors are not necessary. They are written as "trailers" after the actual file body if you don't know the CRC32 and the sizes of the written block when you have to write out the local file header (which precedes the body and has to go "out" before you might even have the chance to have the first byte of the file you are going to be writing). Since the body of a directory will never contain anything (even padding), you always know that the compressed size, the CRC32 and the uncompressed size can all be seto to zeroes. Therefore you know them upfront and you can safely omit the data descriptors even if outputting actual _files_ into the same archive will need them. Data descriptors can be "added" or excluded on a per-entry basis within the same ZIP.

lib/zip_tricks/size_estimator.rb
```diff
@@ -61,4 +61,13 @@ def add_compressed_entry(filename:, uncompressed_size:, compressed_size:, use_da
...
+    udd = !!use_data_descriptor
+    @streamer.add_file_and_write_local_header(filename: "#{dirname}" + "/", crc32: 0, storage_mode: 8,
+      compressed_size: 0, uncompressed_size: 0, use_data_descriptor: udd)
+    @streamer.simulate_write(0)
+    @streamer.write_data_descriptor_for_last_entry if udd
```
julik: See above. The `simulate_write` is also redundant since you don't need to advance the write pointer (`add_file_and_write_local_header` does it for you for the local header)

lib/zip_tricks/streamer.rb
```diff
@@ -158,6 +158,15 @@ def add_stored_entry(filename:, size:, crc32:)
     @out.tell
   end
 
+  # Adds an empty directory to the archive with a size of 0 and permissions of 755.
+  #
+  # @param dirname [String] the name of the directory in the archive
+  def add_empty_directory(dirname:)
```
julik: To make this pitch-perfect YARD you add the same declaration for the return value as for the other methods. In this case it is going to be `  # @return [Fixnum] the offset the output IO is at after writing the entry header`. The benefit of that becomes apparent if you skim through YARD-documented APIs which are automatically rendered for every gem. For example, here is the ZipTricks documentation for a similar method: http://www.rubydoc.info/gems/zip_tricks/4.2.4/ZipTricks%2FStreamer:add_compressed_entry

The `@return` gives the snazzy "return value" specification and the indication that the method returns a Fixnum in the header.

lib/zip_tricks/zip_writer.rb
```diff
@@ -120,11 +126,10 @@ def write_local_file_header(io:, filename:, compressed_size:, uncompressed_size:
   # @param uncompressed_size[Fixnum]  The size of the file once extracted
   # @param crc32[Fixnum] The CRC32 checksum of the file
   # @param mtime[Time]  the modification time to be recorded in the ZIP
-  # @param external_attrs[Fixnum] bit-packed external attributes (defaults to UNIX file with 0644 permissions set)
   # @param gp_flags[Fixnum] bit-packed general purpose flags
   # @return [void]
   def write_central_directory_file_header(io:, local_file_header_location:, gp_flags:, storage_mode:, compressed_size:, uncompressed_size:, mtime:, crc32:, 
-    filename:, external_attrs: DEFAULT_EXTERNAL_ATTRS)
+    filename:)
```
julik: 👍 👍

spec/zip_tricks/streamer_spec.rb
```diff
@@ -102,7 +102,35 @@ def rewind_after(*ios)
...
+      zip.add_empty_directory(dirname: "Tunes")
+    end
+    zip_file.flush
 
+    per_filename = {}
+    Zip::File.open(zip_file.path) do |zip_file|
+      # Handle entries one by one
+      zip_file.each do |entry|
+        # The entry name gets returned with a binary encoding, we have to force it back.
+        per_filename[entry.name] = entry.get_input_stream.read
```
julik: This I'm actually curious about. In my tests, I've ended up with this tidbit:

        if entry.get_input_stream.respond_to?(:read) # For an empty directory, the input stream will be a NullInputStream
          per_filename[entry.name] = entry.get_input_stream.read 
        else
          per_filename[entry.name] = :directory
        end

Haven't you encountered that? (it's a Rubyzip edge case more than anything)

spec/zip_tricks/streamer_spec.rb
```diff
@@ -102,7 +102,35 @@ def rewind_after(*ios)
...
+      end
+    end
+
+    expect(per_filename['Tunes/'].bytesize).to eq(f.size)
```
julik: Since the tempfile is not required you can check this one for nil and be done

spec/zip_tricks/streamer_spec.rb
```diff
@@ -102,7 +102,35 @@ def rewind_after(*ios)
 
     inspect_zip_with_external_tool(zip_file.path)
   end
+  
+  it 'can write and then read an empty directory' do
+    f = Tempfile.new('raw')
+    f.binmode
```
julik: For only the directory you don't even need a source file, since the `add_empty_directory` call does not require anything to be written afterwards, we can proceed straight to the next entry

spec/zip_tricks/streamer_spec.rb
```diff
@@ -129,6 +157,10 @@ def rewind_after(*ios)
       # ...and stored.
       zip.add_stored_entry(filename: 'war-and-peace-raw.txt', size: source_f.size, crc32: crc32)
       zip << source_f.read
+      
+      # Add an empty directory as well.
+      zip.add_empty_directory(dirname: 'Beatles')
+      zip << source_f.read
```
julik: Not necessary - you are just adding filler data to the ZIP file here. Might be useful for stress-testing (or fuzzing) ZIP unarchivers  - see the Reader class for _all the places_ where ZIP reading can and will go wrong - but for this test this is redundant ;-)

spec/zip_tricks/zip_writer_spec.rb
```diff
@@ -212,6 +212,36 @@ def seek_to_start_of_signature(signature)
       expect(br.read_4b).to eq(898921)     # relative offset of local header
       expect(br.read_n(10)).to eq('a-file.txt') # the filename
     end
+
+    it 'writes the file header an entry that contains an empty directory' do
```
julik: 👍  exactly

julik:
Almost! A few more lines and merging.

lib/zip_tricks/size_estimator.rb
bermannoah: 👍

lib/zip_tricks/size_estimator.rb
bermannoah: 👍

lib/zip_tricks/streamer.rb
bermannoah: 👍

spec/zip_tricks/streamer_spec.rb
bermannoah: Hmm. I messed around with this a bit and I can't seem to duplicate the issue -- it never hits the else clause. What are you seeing when it runs into the NullInputStream? 

Only thing I can think of (since it works fine as is) is are we running the same rubyzip version? I've got `1.2.1`. The release notes do mention patching out an issue with `open_buffer` but since this test uses `open`...

spec/zip_tricks/streamer_spec.rb
bermannoah: Wound up having to check this for 0 -- could this be related to the NullInputStream earlier?

PR #14, opened 2017-08-29
Fix quadratic perf problem when uniqing filenames

julik:
We check for unique filenames using a `.any?` on an
Array. It is going to take longer as the array grows,
since `any?` needs to do an array scan. This leads to
a quadratic performance regression when file counts in
the ZIP become large.

Without this patch, writing 40000 fake files into
a null writer takes _minutes_.

### DIFF
--- a/spec/zip_tricks/streamer_spec.rb
+++ b/spec/zip_tricks/streamer_spec.rb
@@ -18,6 +18,24 @@ describe ZipTricks::Streamer do
   def rewind_after(*ios)
     yield.tap { ios.map(&:rewind) }
   end
+  
+  it'is able to deal with an obscene number of files in the ZIP' do
+    d = double()
+    allow(d).to receive(:write_local_file_header)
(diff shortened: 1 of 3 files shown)


spec/zip_tricks/streamer_spec.rb
```diff
@@ -18,6 +18,24 @@
...
+    
+    described_class.open(ZipTricks::NullWriter, writer: d) do |w|
+      40_000.times do |i|
+        w.write_stored_file('%0004d_file' % i) {|body| body << 'w' }
+      end
+    end
+
+    # If there is a performance problem with file counts you will notice
+    # that this test is taking forever
+    expect(true).to eq(true)
```
dsnipe: Why not use a benchmark here and set a high value (just measure on your computer and x10). 
And this thing looks nice: http://philippe.bourgau.net/verify-the-big-o-complexity-of-ruby-code-in-rspec/

PR #16, opened 2017-09-11
Applying rubocop rules

davidenko87:
This change will add rubocop to the library. I've taken the liberty to adjust some of the cops to the needs of this library. Testing the spec files on block length is a bit redundant, same for the method en class length on some of the primary files. Though i suggest to refactor some of the methods, because of the ABC size and the length. 

I've added the `.rubocop_todo.yml` which contains files that need to be fixed.


julik:
Awesome stuff! I'll take it from here ;-)

PR #20, opened 2017-10-26
write the correct crc32

ananthakumaran:
introduced by abb175a75c4eb9b7067d53f4b9dd8420d2a75684

This broke windows excel, though it worked with mac numbers and unzip
also had no problem either.


julik:
Thank you for your PR @ananthakumaran ! This is indeed a bug that slipped through. Can you add a test that covers this please?

ananthakumaran:
It looks like writing meaningful test case for this issue is not straightforward. rubyzip's entry returns the correct CRC irrespective of what we write in the headers.

julik:
Oh I can help.

julik:
@ananthakumaran can you try the branch in #23 and comment there if that works?

PR #25, opened 2017-10-30
Fix unused var in EMPTY_DIRECTORY_EXTERNAL_ATTRS

felixbuenemann:
This fixes a problem in `ZipTricks::ZipWriter::EMPTY_DIRECTORY_EXTERNAL_ATTRS` where `file_type_dir` is defined inside the begin/end block, but `file_type_file` from the definition of `DEFAULT_EXTERNAL_ATTRS` is used.

This causes the specs to fail, so either the current code was right but misleading or the specs test for the wrong behavior.

I don't know which one is right, so I'm submitting this with failing tests.

### DIFF
--- a/lib/zip_tricks/version.rb
+++ b/lib/zip_tricks/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ZipTricks
-  VERSION = '4.5.0'
+  VERSION = '4.5.2'
 end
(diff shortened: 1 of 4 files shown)


julik:
This I need to test by making an archive with some directories, I will get back to this.

julik:
@bermannoah this was your feature - can you verify that with the dir attributes i.e. ArchiveUtility still unzips an empty directory correctly?

bermannoah:
@julik ArchiveUtility and The Unarchiver both unzip an empty dir correctly. :)

julik:
@bermannoah with the proposed changes in this PR applied or without? If the former is the case, can you also change the test assertions, format and merge pls?

PR #26, opened 2017-10-30
Fix duplicate method definition in FakeZipWriter

felixbuenemann:


felixbuenemann:
Btw. there are a bunch of warnings like "assigned but unused variable" or access to undefined instance variables outside of the specs.

Would you accept a PR to fix those warnings?

The warnings can be seen when running `RUBYOPT=-W3 rake`.

julik:
OFC we will. However I would appreciate a .travis.yml change then that would make CI fail when warnings happen.

PR #28, opened 2017-11-04
Adjust CRC32 buffer size to 64 KB and use a buffering wrapper for IO

julik:
Even though the size of the buffer is specific to the CRC32 implementation in zlib, the pattern of buffering writes is actually pretty common - and in the StreamCRC32 objects it is not very declarative. We reimplement it as an write proxy instead, which decouples the buffering stuff and makes it possible to use it in other scenarios as well.

This also adds a benchmark that proves, as correctly stated by @felixbuenemann that the 64KB buffer size _is_ indeed the sweet spot as far as CRC32 is concerned (we intentionally perform 1-byte writes to get the slowest possible throughput and the smallest chunks). This will be beneficial to libraries like XSLX writers which are likely to be writing lots of small chunks in succession (as oposed to archive-from-file situations where `IO.copy_stream` can choose the right chunk size for us).

### DIFF
--- a/lib/zip_tricks/streamer/stored_writer.rb
+++ b/lib/zip_tricks/streamer/stored_writer.rb
@@ -12,7 +12,6 @@ class ZipTricks::Streamer::StoredWriter
 
   def initialize(io)
     @io = ZipTricks::WriteAndTell.new(io)
-    @started_at = @io.tell
     @crc = ZipTricks::WriteBuffer.new(ZipTricks::StreamCRC32.new, CRC32_BUFFER_SIZE)
   end
 
--- a/lib/zip_tricks/streamer/stored_writer.rb
+++ b/lib/zip_tricks/streamer/stored_writer.rb
@@ -4,10 +4,17 @@
 # through it in a CRC32 checksum calculator. Is made to be completely
 # interchangeable with the DeflatedWriter in terms of interface.
 class ZipTricks::Streamer::StoredWriter
+
+  # The amount of bytes we will buffer before computing the intermediate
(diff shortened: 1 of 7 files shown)


felixbuenemann:
This is great work. I was thinking of doing a proper benchmark myself, but didn't yet get around to it.

The sweet spot for CPU intensive operations like this is often related to the L1 cache size, which is 32KB for data on the CPU I was testing on (Intel Core i7-5557U) and most modern desktop and server CPUs.

felixbuenemann:
Btw. there's a typo "Single-bute" vs "Single-byte" in the pasted benchmark results of `bench/buffered_crc32_bench.rb`.

julik:
There is. There is also a thing where we don't need to pre-generate this blob :-P that will be rectified

lib/zip_tricks/streamer/stored_writer.rb
```diff
@@ -1,23 +1,37 @@
...
+  # `bench/buffered_crc32_bench.rb), if that is exceeded Zlib is going
+  # to perform internal CRC combine calls which will make the speed go down again.
+  CRC32_BUFFER_SIZE = 64 * 1024
+
   def initialize(io)
-    @io = io
-    @uncompressed_size = 0
-    @compressed_size = 0
+    @io = ZipTricks::WriteAndTell.new(io)
     @started_at = @io.tell
```
felixbuenemann: You have removed `@started_at` from `#finish` so it can be removed here as well as it isn't referenced anywhere else in the class.

julik: 👍

PR #31, opened 2017-11-21
Rename "testing/" to "qa/"

julik:
primarily to prevent it being picked up by automatic test runners.
Document QA testing in more detail


spec/zip_tricks/zip_writer_spec.rb
```diff
@@ -1,7 +1,6 @@
 # rubocop:disable Layout/CommentIndentation, Metrics/LineLength, Metrics/AbcSize, Style/StringLiterals
 
 require_relative '../spec_helper'
-require_relative '../../testing/support'
```
davidenko87: Where do you include the new file?

julik: Tests pass without it, so nowhere

PR #36, opened 2018-01-24
Rubocop integration

linkyndy:
I (tried to) integrate Rubocop here through our gem. Fixed some of the offenses, but there are still some left that you're more entitled to decide upon. Let me know if something is not working for you :)


lib/zip_tricks/file_reader.rb
```diff
@@ -669,6 +669,7 @@ def num_files_and_central_directory_offset_zip64(io, zip64_end_of_cdir_location)
   # central directory record upfront (all the fixed fields + at most 0xFFFF
   # bytes of the archive comment)
   MAX_END_OF_CENTRAL_DIRECTORY_RECORD_SIZE =
+    # rubocop:disable Layout/MultilineOperationIndentation
```
julik: What is the proposed Rubocop formatting for these constants defined with a block?

julik:
❗️ BTW what's the idea of approving with CI failing still? That's not how it works :-) that's not how anything of this works....

lib/zip_tricks/file_reader.rb
linkyndy: Well:

```
1+
2+
...
```

So, no indentation.

linkyndy:
@julik I wasn't seeking approval yet, but more your contribution to fixing the offenses, since you know best what's supposed to be there (see my PR description) 😀

julik:
Ah, so that's how them tables turn now. Ok ok, but I won't get to this until the end of this week I am afraid :-(

julik:
@linkyndy I think that will do. I have applied 3 exceptions in total:

* The QA scripts generate large files and are messy, they are also not used anywhere else - I think using global variables in them is justified
* I absolutely hate the first-argument-on-next-line rules. Can't help it. Can we maybe reconsider them? This is a personal preference.

The rest should be appeased.

linkyndy:
Great job! I agree some projects can have exceptions from the default cops list, so that's fine. Regarding the line breaks/first arguments, can you point me to an example? I didn't see one while I was skimming through the changes.

.rubocop.yml
```diff
@@ -1,141 +1,10 @@
...
-
-Lint/AssignmentInCondition:
+inherit_gem:
+  wetransfer_style: ruby/default.yml
+Layout/FirstMethodArgumentLineBreak:
   Enabled: false
-
-# We use % all over and nothing ever came out of it
-Style/FormatString:
+Layout/FirstMethodParameterLineBreak:
```
davidenko87: can we add these two to the general style gem?

linkyndy: Can you point me to an actual example of this in the PR? Not that it is dependent on including it in the style gem, but I want to understand better what's the problem.

davidenko87: For example, in `size_estimator.rb` and `steamer.rb` we place the arguments on the next line. Inside `size_estimator_spec.rb` and `file_reader_spec.rb` these are on one line or the first line is in the same as the method.

linkyndy: Well, I think the problem is occurring only for `size_estimator_spec.rb`, where I tend to agree with rubocop. That way of aligning arguments is very specific to Python. But anyways, that's my taste, so if you're against it, let's remove the cops from `wetransfer_style`. The aim is to enforce stuff while keeping everyone happy (within reasonable boundaries) :)

julik:
@davidenko87 fair point

linkyndy:
So what do you guys propose?

linkyndy:
Well, removing the cops will allow multiple styles, so it cancels your first statement. How do you want to enforce that style?

davidenko87:
Well, we could edit the linter so it matches our needs.

linkyndy:
Wouldn't go that far...

davidenko87:
What is your proposal to have one style in this case?

linkyndy:
Either keep the cops, either remove them and allow people different styles. We are anyway disagreeing on which style to use. Maybe it's not that bad to be a bit flexible and allow everyone the freedom to use whatever style they want.

julik:
@linkyndy @davidenko87 I think we are in the clear now. I reverted the changes for first-arg-on-next-line to have consistent style and applied a reformat so that args/kwargs get realigned. I think this can be merged 🎉

davidenko87:
i agree!

julik:
There is a couple exceptions that I missed when I was removing reminder comments, I'll deal with them.

julik:
@linkyndy aaaalmost there. Can you take a look how we should deal with codeclimate? If it is going to keep failing like this we should remove it since we will be running rubocop for each CI cycle anyway.

julik:
Killed Codeclimate for now, we can revisit later ;-)

linkyndy:
Yep, looks good! Regarding Codeclimate, never actually opened it 😛 But indeed, we should rethink so it doesn't stay in our way. I like it as a tool, so would be nice to manage to keep it somehow.

PR #42, opened 2018-04-10
Test with Ruby 2.5 on Travis CI

aried3r:
Personally I'd remove official support for 2.1 and 2.2 which are EOL for the next major version.


julik:
Unfortunately we now have libraries like zipline that depend on us, and if we stop testing on 2.1/2.2 there is a chance we will suddenly use a piece of syntax which is unsupported. So I'd rather keep them in the config.

2.5 is certainly a good one to add - but you might need to do a bundler update, the Travis task fails atm.

aried3r:
I'll try to fix 2.5.

Regarding Ruby 2.1/2.2, at least zipline has a restriction on [zip_tricks < 5](https://github.com/fringd/zipline/blob/32f1f04ed23a65ef804b3af4d6e52d0fe0bdaaf5/zipline.gemspec#L18), and dropping Ruby versions is a breaking change and thus should be a major version change.

There are [other dependents](https://github.com/WeTransfer/zip_tricks/network/dependents) but only zipline and [xlsxstream](https://github.com/felixbuenemann/xlsxtream/blob/844f0960a7bb1f7525f00300bddf01f92e390794/xlsxtream.gemspec#L23) seem to be used a bit (judging from GitHub stars).

I'm in no rush to drop these Ruby versions, but I also think you shouldn't be afraid to bump the major version number and drop these versions. :)
Old versions of zip_tricks will continue to work anyway.

Anyway, thanks a lot for this gem!

aried3r:
Seems to work now. Do you want me to squash or are you gonna do it using the GitHub merge options?

julik:
I'll squash on merge. Thanks a lot for the changes!

Issue #41, opened 2018-04-05
Specify modification date when adding entries to the Streamer


RTrampov:
As I mentioned in #40 we had another requirement for which there was no readily available solution.

We save file modification dates in the database and wanted to preserve them when archiving files, so the modification date is the same in the archive as it is on the server. Here's an example:
```ruby
ZipTricks::Streamer.open(output_stream) do |streamer|
  streamer.write_stored_file('text-file.txt', modification_date: Time.now.utc) do |entry|
    # ...
  end
  streamer.write_deflated_file('picture.png', modification_date: Time.now.utc) do |entry|
    # ...
  end
end
```

Again - if you are interested in this I can try and open a pull request.

julik:
Thanks for the proposal! 

Yes, that could clearly be useful. Since we already set creation/modification time extra fields for all entries output, would it be useful to set the creation time/access times as well?

RTrampov:
Well our requirement is for modification time only but I guess create/access time could be useful to someone else.

That would require modification of `ZipTricks::ZipWriter#timestamp_extra` right? As it currently only accepts and sets the modification time?

julik:
Correct. And we will need to add an argument to all the related methods (in Streamer and Writer)

RTrampov:
For the time being I can open a PR about the modification time as it requires less amount of changes and maybe add create/access time later when I have the time?

julik:
Yep, go for it!

julik:
Fixed anf merged, thanks again!

PR #43, opened 2018-04-12
Allow to specify modification time when adding entries to streamer

RTrampov:
Resolving #41. 

Added optional `modification_time` argument to all `add_*` and `write_*` streamer methods which is then passed to `add_file_and_write_local_header`.


julik:
LGTM, thank you for the changes!

PR #57, opened 2019-05-29
Optimize CRC32 allocations for IOs

julik:
We can reuse a string buffer across multiple reads, thereby
reducing allocations and GC churn. Also we can raise the
buffer size as reading the file is usually more expensive than
computing the CRC. And to top it off on recent Rubies we can
pre-cap our String to the right size and it will not have to
be resized on every alteration, sparing cycles. It uses two
semi-obscure features:

**Capacity preallocation**

The constructor for `String` accepts an optional `capacity`
keyword argument, since
https://bugs.ruby-lang.org/issues/12024 The argument can be
used to pre-size the String to the specific number of bytes.
If the string gets mutated, and the new string contents is not
larger than the `capacity` of the string, the string will not
need resizing - meaning we never need to realloc for the
string contents. We must run on older Rubies too so that
optimization is conditional.

**IO#read with the buffer argument**

All `IO#read` calls accept an optional `outbuf` argument which
should be a String. IO then replaces the contents of that
string instead of allocating a new one, and returns it. This
makes it possible to reuse one and the same string object
throughout multiple `read` calls. See
https://apidock.com/ruby/IO/read for more. Since we only read
as many bytes as we set in `capacity`, we won't be resizing
the string either.

We use a larger buffer size as this method has different
characteristics than the one we use for stream-writes from
the Streamer.

**Benchmarks**

https://gist.github.com/julik/ada7229c8d85290aad3c72572c368f01

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,3 +8,4 @@ Style/GlobalVars:
   Exclude:
       - qa/*.rb
       - spec/spec_helper.rb
+      - spec/support/zip_inspection.rb
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.7.4
+
+* Use a single fixed capacity string in StreamCRC32.from_io to avoid unnecessary allocations
+
 ## 4.7.3
 
 * Fix RemoteUncap#request_object_size to function correctly
--- a/lib/zip_tricks/stream_crc32.rb
+++ b/lib/zip_tricks/stream_crc32.rb
@@ -2,13 +2,26 @@
 
 # A simple stateful class for keeping track of a CRC32 value through multiple writes
 class ZipTricks::StreamCRC32
+  STRINGS_HAVE_CAPACITY_SUPPORT = begin
+    String.new('', capacity: 1)
+    true
+  rescue ArgumentError
+    false
+  end
+  CRC_BUF_SIZE = 1024 * 512
+  private_constant :STRINGS_HAVE_CAPACITY_SUPPORT, :CRC_BUF_SIZE
+
   # Compute a CRC32 value from an IO object. The object should respond to `read` and `eof?`
   #
   # @param io[IO] the IO to read the data from
   # @return [Fixnum] the computed CRC32 value
   def self.from_io(io)
+    # If we can specify the string capacity upfront we will not have to resize
+    # the string during operation. This saves time but is only available on
+    # recent Ruby 2.x versions.
+    blob = STRINGS_HAVE_CAPACITY_SUPPORT ? String.new('', capacity: CRC_BUF_SIZE) : String.new('')
     crc = new
-    crc << io.read(1024 * 512) until io.eof?
+    crc << io.read(CRC_BUF_SIZE, blob) until io.eof?
     crc.to_i
   end
 
--- a/lib/zip_tricks/version.rb
+++ b/lib/zip_tricks/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ZipTricks
-  VERSION = '4.7.3'
+  VERSION = '4.7.4'
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -7,71 +7,11 @@ require 'digest'
 require 'fileutils'
 require 'shellwords'
 require 'zip'
-require 'delegate'
 
-class ReadMonitor < SimpleDelegator
-  def read(*)
-    super.tap do
-      @num_reads ||= 0
-      @num_reads += 1
-    end
-  end
-
-  def num_reads
-    @num_reads || 0
-  end
-end
-
-class ManagedTempfile < Tempfile
-  @@managed_tempfiles = []
-
-  def initialize(*)
-    super
-    @@managed_tempfiles << self
-  end
-
-  def self.prune!
-    @@managed_tempfiles.each do |tf|
-      begin
-        tf.close
-        tf.unlink
-      rescue
-        nil
-      end
-    end
-    @@managed_tempfiles.clear
-  end
-end
-
-module ZipInspection
-  def inspect_zip_with_external_tool(path_to_zip)
-    zipinfo_path = 'zipinfo'
-    $zip_inspection_buf ||= StringIO.new
-    $zip_inspection_buf.puts "\n"
-    # The only way to get at the RSpec example without using the block argument
-    $zip_inspection_buf.puts "Inspecting ZIP output of #{inspect}."
-    $zip_inspection_buf.puts 'Be aware that the zipinfo version on OSX is too \
-                              old to deal with Zip64.'
-    escaped_cmd = Shellwords.join([zipinfo_path, '-tlhvz', path_to_zip])
-    $zip_inspection_buf.puts `#{escaped_cmd}`
-  end
-
-  def open_with_external_app(app_path, path_to_zip, skip_if_missing)
-    bin_exists = File.exist?(app_path)
-    skip "This system does not have #{File.basename(app_path)}" if skip_if_missing && !bin_exists
-    return unless bin_exists
-    `#{Shellwords.join([app_path, path_to_zip])}`
-  end
-
-  def open_zip_with_archive_utility(path_to_zip, skip_if_missing: false)
-    # ArchiveUtility sometimes puts the stuff it unarchives in ~/Downloads etc. so do
-    # not perform any checks on the files since we do not really know where they are on disk.
-    # Visual inspection should show whether the unarchiving is handled correctly.
-    au_path = '/System/Library/CoreServices/Applications/Archive Utility.app/ \
-              Contents/MacOS/Archive Utility'
-    open_with_external_app(au_path, path_to_zip, skip_if_missing)
-  end
-end
+require_relative 'support/read_monitor'
+require_relative 'support/managed_tempfile'
+require_relative 'support/zip_inspection'
+require_relative 'support/allocate_under_matcher'
 
 RSpec.configure do |config|
   config.include ZipInspection
--- a/spec/support/allocate_under_matcher.rb
+++ b/spec/support/allocate_under_matcher.rb
@@ -1,0 +1,27 @@
+require 'allocation_stats'
+RSpec::Matchers.define :allocate_under do |expected|
+  match do |actual|
+    @trace = actual.is_a?(Proc) ? AllocationStats.trace(&actual) : actual
+    @trace.new_allocations.size < expected
+  end
+
+  def objects
+    self
+  end
+
+  def supports_block_expectations?
+    true
+  end
+
+  def output_trace_info(trace)
+    trace.allocations(alias_paths: true).group_by(:sourcefile, :sourceline, :class).to_text
+  end
+
+  failure_message do |_actual|
+    "expected under #{expected} objects to be allocated; got #{@trace.new_allocations.size}:\n\n" << output_trace_info(@trace)
+  end
+
+  description do
+    "allocates under #{expected} objects"
+  end
+end
--- a/spec/support/managed_tempfile.rb
+++ b/spec/support/managed_tempfile.rb
@@ -1,0 +1,20 @@
+class ManagedTempfile < Tempfile
+  @@managed_tempfiles = []
+
+  def initialize(*)
+    super
+    @@managed_tempfiles << self
+  end
+
+  def self.prune!
+    @@managed_tempfiles.each do |tf|
+      begin
+        tf.close
+        tf.unlink
+      rescue
+        nil
+      end
+    end
+    @@managed_tempfiles.clear
+  end
+end
--- a/spec/support/read_monitor.rb
+++ b/spec/support/read_monitor.rb
@@ -1,0 +1,14 @@
+require 'delegate'
+
+class ReadMonitor < SimpleDelegator
+  def read(*)
+    super.tap do
+      @num_reads ||= 0
+      @num_reads += 1
+    end
+  end
+
+  def num_reads
+    @num_reads || 0
+  end
+end
--- a/spec/support/zip_inspection.rb
+++ b/spec/support/zip_inspection.rb
@@ -1,0 +1,29 @@
+module ZipInspection
+  def inspect_zip_with_external_tool(path_to_zip)
+    zipinfo_path = 'zipinfo'
+    $zip_inspection_buf ||= StringIO.new
+    $zip_inspection_buf.puts "\n"
+    # The only way to get at the RSpec example without using the block argument
+    $zip_inspection_buf.puts "Inspecting ZIP output of #{inspect}."
+    $zip_inspection_buf.puts 'Be aware that the zipinfo version on OSX is too \
+                              old to deal with Zip64.'
+    escaped_cmd = Shellwords.join([zipinfo_path, '-tlhvz', path_to_zip])
+    $zip_inspection_buf.puts `#{escaped_cmd}`
+  end
+
+  def open_with_external_app(app_path, path_to_zip, skip_if_missing)
+    bin_exists = File.exist?(app_path)
+    skip "This system does not have #{File.basename(app_path)}" if skip_if_missing && !bin_exists
+    return unless bin_exists
+    `#{Shellwords.join([app_path, path_to_zip])}`
+  end
+
+  def open_zip_with_archive_utility(path_to_zip, skip_if_missing: false)
+    # ArchiveUtility sometimes puts the stuff it unarchives in ~/Downloads etc. so do
+    # not perform any checks on the files since we do not really know where they are on disk.
+    # Visual inspection should show whether the unarchiving is handled correctly.
+    au_path = '/System/Library/CoreServices/Applications/Archive Utility.app/ \
+              Contents/MacOS/Archive Utility'
+    open_with_external_app(au_path, path_to_zip, skip_if_missing)
+  end
+end
--- a/spec/zip_tricks/stream_crc32_spec.rb
+++ b/spec/zip_tricks/stream_crc32_spec.rb
@@ -11,7 +11,12 @@ describe ZipTricks::StreamCRC32 do
 
   it 'when computing the CRC32 from an IO only allocates one String' do
     raw = StringIO.new(Random.new.bytes(45 * 1024 * 1024))
-    expect { described_class.from_io(raw) }.to allocate_under(3).objects
+    # The number of objects allocated depends on the MRI version, but in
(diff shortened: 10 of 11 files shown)


julik:
cc @felixbuenemann since we were busy with CRC-related buffers last ;-)

felixbuenemann:
@julik What were your benchmark results? Maybe you could post them as a comment on the gist.

julik:
Posted, also posting here for posterity ;-)

Results (2.5.1 x86_64-darwin18):

```
Allocations (512k)
       sourcefile         sourceline  class_path       method_id        memsize  class
------------------------  ----------  ----------  --------------------  -------  ------
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 30  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<RUBYLIBDIR>/delegate.rb         349  IO          read                   524329  String
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 27  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 26  Object      blockwise_crc32_512k       40  Array
<PWD>/crc32_bm.rb                 25  Object      blockwise_crc32_512k       40  String
<PWD>/crc32_bm.rb                 24  Object      blockwise_crc32_512k       40  Array
Allocations (512k with string reuse)
   sourcefile      sourceline  class_path            method_id             memsize  class
-----------------  ----------  ----------  ------------------------------  -------  ------
<PWD>/crc32_bm.rb          54  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          51  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          50  Object      blockwise_crc32_512k_reuse_str       40  Array
<PWD>/crc32_bm.rb          49  String      initialize                          192  Hash
<PWD>/crc32_bm.rb          49  Class       new                              524329  String
<PWD>/crc32_bm.rb          49  Object      blockwise_crc32_512k_reuse_str      192  Hash
<PWD>/crc32_bm.rb          49  Object      blockwise_crc32_512k_reuse_str       40  String
<PWD>/crc32_bm.rb          48  Object      blockwise_crc32_512k_reuse_str       40  String
<PWD>/crc32_bm.rb          47  Object      blockwise_crc32_512k_reuse_str       40  Array
Warming up --------------------------------------
blockwise_crc32_64k           
                        12.000  i/100ms
blockwise_crc32_512k          
                        22.000  i/100ms
blockwise_crc32_64k_reuse_str 
                        15.000  i/100ms
blockwise_crc32_512k_reuse_str
                        26.000  i/100ms
Calculating -------------------------------------
blockwise_crc32_64k           
                        136.417  (± 5.1%) i/s -    684.000  in   5.027239s
blockwise_crc32_512k          
                        230.368  (± 3.5%) i/s -      1.166k in   5.067632s
blockwise_crc32_64k_reuse_str 
                        159.990  (± 1.3%) i/s -    810.000  in   5.063623s
blockwise_crc32_512k_reuse_str
                        267.094  (± 1.1%) i/s -      1.352k in   5.062690s

Comparison:
blockwise_crc32_512k_reuse_str:      267.1 i/s
blockwise_crc32_512k          :      230.4 i/s - 1.16x  slower
blockwise_crc32_64k_reuse_str :      160.0 i/s - 1.67x  slower
blockwise_crc32_64k           :      136.4 i/s - 1.96x  slower
```

felixbuenemann:
@julik I think it would be worthwhile to re-use the CRC buffer.

Allocating such a large string over and over puts quite some strain on the GC.

If I reuse the buffer in the benchmark I get pretty much constant RSS regardless of buffer size, while otherwise RSS roughly grows by the same factor as buffer size.

Benchmark 20s only `:blockwise_crc32_64k_reuse_str`: 40 MB RSS
Benchmark 20s only `:blockwise_crc32_64k_reuse_str_reuse_buf`: 24 MB RS
Benchmark 20s only `:blockwise_crc32_512k_reuse_str`: 165 MB RSS
Benchmark 20s only `:blockwise_crc32_512k_reuse_str_reuse_buf`: 24 MB RSS

(If you disable GC, the benchmark will allocate multiple gigabytes of memory if the buffers is not re-used between method calls.)

In the benchmark script I used the following code to get a thread-local buffer:

```ruby
  blob = if Thread.current.thread_variable?(:crc_buf_512k)
    Thread.current.thread_variable_get(:crc_buf_512k)
  else
    Thread.current.thread_variable_set(:crc_buf_512k, String.new('', capacity: 524_288))
  end
```

However in the library a class-level memoized getter should work:

```ruby
class ZipTricks::StreamCRC32
  def self.crc_buffer
    @crc_buffer ||= begin
      String.new('', capacity: 524_288)
    rescue ArgumentError
      String.new('')
    end
  end
end
```

Btw. I don't think that preallocating the string is that useful. I couldn't find any performance difference when running the benchmark with the `capacity: n` removed.

Since it will already read the IO in fixed chunks, it will just reallocate the buffer once on the first read.

This gets even less relevant, if the buffer is reused across instances.

felixbuenemann:
@julik I came up with the following changes. Let me know what you think!

```diff
diff --git a/lib/zip_tricks/stream_crc32.rb b/lib/zip_tricks/stream_crc32.rb
index 3e47d75..2d50fc3 100644
--- a/lib/zip_tricks/stream_crc32.rb
+++ b/lib/zip_tricks/stream_crc32.rb
@@ -2,27 +2,28 @@
 
 # A simple stateful class for keeping track of a CRC32 value through multiple writes
 class ZipTricks::StreamCRC32
-  STRINGS_HAVE_CAPACITY_SUPPORT = begin
-    String.new('', capacity: 1)
-    true
-  rescue ArgumentError
-    false
-  end
   CRC_BUF_SIZE = 1024 * 512
-  private_constant :STRINGS_HAVE_CAPACITY_SUPPORT, :CRC_BUF_SIZE
+  private_constant :CRC_BUF_SIZE
+
+  def self.crc_buf
+    @crc_buf ||= begin
+      String.new('', capacity: CRC_BUF_SIZE)
+    rescue ArgumentError
+      String.new('')
+    end
+  end
+  private_class_method :crc_buf
 
   # Compute a CRC32 value from an IO object. The object should respond to `read` and `eof?`
   #
   # @param io[IO] the IO to read the data from
   # @return [Fixnum] the computed CRC32 value
   def self.from_io(io)
-    # If we can specify the string capacity upfront we will not have to resize
-    # the string during operation. This saves time but is only available on
-    # recent Ruby 2.x versions.
-    blob = STRINGS_HAVE_CAPACITY_SUPPORT ? String.new('', capacity: CRC_BUF_SIZE) : String.new('')
     crc = new
-    crc << io.read(CRC_BUF_SIZE, blob) until io.eof?
+    crc << io.read(CRC_BUF_SIZE, crc_buf) until io.eof?
     crc.to_i
+  ensure
+    crc_buf.clear
   end
 
   # Creates a new streaming CRC32 calculator
```

The `crc_buf.clear` is not needed and causes slightly higher memory usage, but it's probably a good idea to leave no file contents in a shared buffer.

julik:
Sorry I went a bit trigger happy with merging. Reusing the buffer is a good idea when there are no threads. When there are threads (and where we use ZipTricks there are threads, many...) you get into managing the thread-locals for that buffer and allocating it again and again. Then you need a "connection pool" of sorts of these shared string buffers, or a ring queue.

```
BUFFER_POOL = Queue.new

def self.from_io(io)
  buf = BUFFER_POOL.pop(non_block=true)
  crc = new
  crc << io.read(512*1024, buf) until io.eof?
  crc.to_i
rescue ThreadError
  # If many many many from_io calls happen at the same time
  # we will be leaking those strings
  new_buf = String.new('', capacity: 512*1024) rescue String.new('')
  BUFFER_POOL.push(new_buf)
  retry
end
```

To implement this cleanly one would need to take connection_pool on as a dependency.

Alternative: we could add an optional argument for a buffer so that it can be reused manually (does feel like C/Go though).

felixbuenemann:
Why do you need a connection pool? The class instance variable is already thread local, so each thread gets its own buffer. 

Do you have so many threads that a 512k buffer per thread is prohibitive?

julik:
Not that many threads, but we will need to allocate a new one for each thread we spin up - or, even more importantly, one per _fiber_ we allocate because thread locals are ultimately fiber locals.

> The class instance variable is already thread local

Are they though? If i put together an example like this the `@mut` is nicely shared between all the threads and they clobber each other's input...

```ruby
class Some
  @mut = String.new('')
  def self.mutate_buf
    random_char = ('a'..'z').to_a.sample
    @mut.replace(random_char)
    sleep(rand / 8) # Allow another thread to intervene here
    puts "Replaced @mut with #{random_char} and the value is now #{@mut}"
  end
end

(1..120).map do
  Thread.new do
    Some.mutate_buf 
  end
end.map(&:join)
```

felixbuenemann:
Thread locals and fiber locals are not the same. Fiber locals are accessed by Thread.current[] while thread locals use thread_variable_get and set. 

My understanding was that creating a memoized instance var from a class method is allocated on the eigenclass and is thread local while using @@ instance vars is shared and not thread safe. 

Your example defined the instance var static at the class level instead of from a class method, so I'm not sure thats a valid test.

Its totally possible that my understanding of @ and @@ ivars at the class level is wrong, but it should be easy to check.

felixbuenemann:
Looks like I should read up on the ivar vs cvar difference in class context again, memoizing to an ivar from a class method is definitely not local to the thread.

So that code in my diff would have to be rewritten to explicitly use `thread_variable_get` and `thread_variable_set`.

felixbuenemann:
@julik Here's an updated diff to use a thread (but not fiber) local buffer:

```diff
diff --git a/lib/zip_tricks/stream_crc32.rb b/lib/zip_tricks/stream_crc32.rb
index 3e47d75..6dad275 100644
--- a/lib/zip_tricks/stream_crc32.rb
+++ b/lib/zip_tricks/stream_crc32.rb
@@ -11,18 +11,26 @@ class ZipTricks::StreamCRC32
   CRC_BUF_SIZE = 1024 * 512
   private_constant :STRINGS_HAVE_CAPACITY_SUPPORT, :CRC_BUF_SIZE
 
+  def self.crc_buf
+    if Thread.current.thread_variable?(:zip_tricks_crc_buf)
+      Thread.current.thread_variable_get(:zip_tricks_crc_buf)
+    else
+      blob = STRINGS_HAVE_CAPACITY_SUPPORT ? String.new('', capacity: CRC_BUF_SIZE) : String.new('')
+      Thread.current.thread_variable_set(:zip_tricks_crc_buf, blob)
+    end
+  end
+  private_class_method :crc_buf
+
   # Compute a CRC32 value from an IO object. The object should respond to `read` and `eof?`
   #
   # @param io[IO] the IO to read the data from
   # @return [Fixnum] the computed CRC32 value
   def self.from_io(io)
-    # If we can specify the string capacity upfront we will not have to resize
-    # the string during operation. This saves time but is only available on
-    # recent Ruby 2.x versions.
-    blob = STRINGS_HAVE_CAPACITY_SUPPORT ? String.new('', capacity: CRC_BUF_SIZE) : String.new('')
     crc = new
-    crc << io.read(CRC_BUF_SIZE, blob) until io.eof?
+    crc << io.read(CRC_BUF_SIZE, crc_buf) until io.eof?
     crc.to_i
+  ensure
+    crc_buf.clear
   end
 
   # Creates a new streaming CRC32 calculator
```

julik:
This will probably be an issue with socketry/async. A `read` call in there can be happening inside a task/fiber, and if fibers will be reusing a buffer there it is going to be clobbered.

What options do we have? Can we make an optional argument for the reusable buffer? Or reduce the buffer size so that we can safely allocate it within the method.

felixbuenemann:
It my understanding that fibers use cooperative multitasking, so they don't really run at the same time, which means sharing the buffer should be fine.

A buffer argument would work, but as you mentioned it makes the API a bit uglier to use. Then again thats the same way the Ruby IO APIs do it. It could also be an optional argument and allocate it inside the function if not specified.

The nice thing about the buffer argument is that the buffer can cleanly be collected be collected by GC once it's no longer needed while the class level buffer would occupy memory for the whole runtime of a thread.

I'm leaning towards the simpler solution of an optional buffer argument. It would introduce no breaking changes to the API and would still allow to optimize memory usage and GC churn if that is an issue in the particular app.

julik:
> I'm leaning towards the simpler solution of an optional buffer argument.

+1. Optional buffer argument or default method-local String. Can you make a PR?

PR #59, opened 2019-05-31
Fix broken spec due to bin vs text mode compare

felixbuenemann:
This fixes a spec that was broken due to comparing the beginning of a text file with an UTF8-BOM with contents read in text mode, which strips the BOM bytes. Always reading in binmode avoids the error.

Without the fix the spec failed with the following error:

```ruby
  2) ZipTricks::Streamer creates an archive with data descriptors that can be opened by Rubyzip, with a small number of very tiny text files
     # https://github.com/rubyzip/rubyzip/issues/295
     Failure/Error: expect(readback[0..10]).to eq(File.read(__dir__ + '/war-and-peace.txt')[0..10])

       expected: "The Projec"
            got: "\xEF\xBB\xBFThe Proj"

       (compared using ==)
     # ./spec/zip_tricks/streamer_spec.rb:461:in `block (3 levels) in <top (required)>'
```

### DIFF
--- a/spec/zip_tricks/streamer_spec.rb
+++ b/spec/zip_tricks/streamer_spec.rb
@@ -431,23 +431,21 @@ describe ZipTricks::Streamer do
   it 'creates an archive with data descriptors that can be opened by Rubyzip, with a small number of very tiny text files' do
     tf = ManagedTempfile.new('zip')
     described_class.open(tf) do |zip|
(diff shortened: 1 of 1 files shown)


julik:
Oh wow! I didn't expect an institution as respectable as the Gutenberg project to introduce BOMs into their material. 👍

PR #60, opened 2019-08-08
Add directory/file clobber check

julik:
We have been having a number of situations where
an implicitly created directory in longer paths
would "occupy" a path component that would then
be taken up by a file. With some applications it
is simply not possible to extract such an archive,
and also the filesystem structure it would create
is not really defined.

To avoid this we should verify the paths added
during the archive creation - and we need to do
it when every entry gets added to the internal
entry list, at the earliest possible opportunity.

It also shows that our uniquify calls to avoid
duplicate filenames lead to ambiguous situations,
so we make this optional with a default of "true".

We keep the option enabled by default to make sure
we do not break zipline which relies on us doing
the uniqueness fixing for them (we asked for it).

So even though we can fix it in our own output,
it is not reasonable to expect all of our users to
do the same, _and_ there are guarantees of semver.

### DIFF
--- a/lib/zip_tricks/path_set.rb
+++ b/lib/zip_tricks/path_set.rb
@@ -1,0 +1,148 @@
+# rubocop:disable Layout/IndentHeredoc
+
+# A ZIP archive contains a flat list of entries. These entries can implicitly
+# create directories when the archive is expanded. For example, an entry with
+# the filename of "some folder/file.docx" will make the unarchiving application
+# create a directory called "some folder" automatically, and then deposit the
+# file "file.docx" in that directory. These "implicit" directories can be
+# arbitrarily nested, and create a tree structure of directories. That structure
+# however is implicit as the archive contains a flat list.
+#
+# This creates opportunities for conflicts. For example, imagine the following
+# structure:
+#
+# * `something/` - specifies an empty directory with the name "something"
+# * `something` - specifies a file, creates a conflict
+#
+# This can be prevented with filename uniqueness checks. It does get funkier however
+# as the rabbit hole goes down:
+#
+# * `dir/subdir/another_subdir/yet_another_subdir/file.bin` - declares a file and directories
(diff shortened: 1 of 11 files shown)


lib/zip_tricks/path_set.rb
```diff
@@ -0,0 +1,148 @@
...
+  # as this would cause conflict.
+  #
+  # The operation also adds all the containing directories for the file, so
+  #    add_file_path("dir/dir2/file.doc")
+  # will add "dir" and "dir/dir2" as directories, "dir/dir2/dir3".
+  #
+  # @param file_path[String] the path to the directory to add
+  # @return [void]
+  def add_file_path(file_path)
+    if @known_files.include?(file_path)
```
arnoFleming: Before checking whether the set `include?` anything, feel free to check if the set's [`add?`](https://ruby-doc.org/stdlib-2.6.3/libdoc/set/rdoc/Set.html#method-i-add-3F) might do exactly what you want. TL;DR: Adding a thing if it isn't in the set yet, returning `nil` if it's a duplicate.

arnoFleming: I used it like [this](https://github.com/WeTransfer/wetransfer_ruby_sdk/pull/59/files#diff-7f6c1b9db8a586878d187e96a2fe3fc2R47) to check for unique file names in the ruby sdk

julik: Good point. In this case however I would end up with a negative `if`:  `if !@set.add?(thing)` which does not communicate intent all that well. It would be, in theory, very welcome to make the PathSet "self-reverting" if a Conflict is raised (do not add either the element that caused the conflict itself, nor the implicit directories it caused to get inserted), and then `add?` would make more sense. The problem however is that even with us some ZIPs contain up to multiple thousands of entries. And for each of these entries we would either need to create backup copies of both sets, or implement some kind of "undo stack" with checkpoints - which would remove only the freshly added elements if a failure occurs. Probably I'm going to do that in the next iteration.

PR #61, opened 2019-10-31
Adopt Hippocratic license

julik:
As @bermannoah noticed in the format_parser pull request it is a good idea for us to adopt a more restrictive license.

This makes the license zip_tricks is under more restrictive than it used to be. To limit impact on downstream dependencies such as xlsxtream I opted to increment the major version of the gem so that its update will be more visible and the dependents can elect to stay on 4.x version tree.

The 4.x will stay under the basic MIT license so that we do not suddenly de-license current users of the software.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 5.0.0
 
 * Adopt [Hippocratic license v. 1.2](https://firstdonoharm.dev/version/1/2/license.html)
+  Note that this might make the license conditions unacceptable for your project. If that is the case,
+  you can use the 4.x branch of the library which stays under the original, exact MIT license.
 
 ## 4.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.0
+
+* Adopt [Hippocratic license v. 1.2](https://firstdonoharm.dev/version/1/2/license.html)
+
 ## 4.8.0
 
 * Make sure that when directories clobber files and vice versa we raise a clear error. Add `PathSet` which keeps track of entries
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,20 +1,17 @@
-Copyright (c) 2019 WeTransfer
+Copyright 2019 WeTransfer
 
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
-"Software"), to deal in the Software without restriction, including
-without limitation the rights to use, copy, modify, merge, publish,
-distribute, sublicense, and/or sell copies of the Software, and to
-permit persons to whom the Software is furnished to do so, subject to
-the following conditions:
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
-The above copyright notice and this permission notice shall be
-included in all copies or substantial portions of the Software.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
-OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
-WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+* The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+* No Harm: The software may not be used by anyone for systems or activities that actively and knowingly endanger, harm, or otherwise threaten the physical, mental, economic, or general well-being of other individuals or groups, in violation of the United Nations Universal Declaration of Human Rights (https://www.un.org/en/universal-declaration-human-rights/).
+
+* Services: If the Software is used to provide a service to others, the licensee shall, as a condition of use, require those others not to use the service in any way that violates the No Harm clause above.
+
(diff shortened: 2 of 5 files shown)


julik:
@felixbuenemann For context: Noah filed this PR initially https://github.com/WeTransfer/format_parser/pull/142 and I didn't think this would fly, but apparently we can - given our company obligations and concerns - adopt this license for most of our open-source libraries. I do agree with Noah that if we can do it we definitely should.

I would love to have your angle as the maintainer of a key downstream dependent in this case, outside of the ethical position which might be at play. Worst case scenario we can keep the 4.x branch alive for some time with bugfixes if the proposition is not viable on your end.

felixbuenemann:
@julik Do you have any information on the legal implications for downstream projects?

I'm no advocate, but since this license is based on MIT and does not prohibit the use in projects that have a license that also requires ethical use, I don't think there is an issue.

It is my understanding that users of a downstream library like xlsxtream need to comply with the license of all its dependencies as well, so as long as _I_ as the library author don't use it for non-ethic things it should be fine.

___

That said I'm skeptical of the real world impact of such a license.

Governments that violate human rights tend to care very little wether they violate copyrights or software licenses and you would have a very hard time bringing up a legal case against them and the same goes for criminal organizations.

It would be interesting to see, if there are known cases where such a license was able to actively prevent violations against human rights.

I think using such a license is a bit like selling a gun with a plate that says "don't use to harm humans". While this would be well intended, the impact would be zero and you would be far better of by preventing the sale of guns to the general public.

So in short I think a license can only help you, if the users of your software are willing to work within the boundaries of law, which already should prevent violations against human rights.

___

I think if we want to promote ethical software, we as engineers need to actively say no to implementing features that for example invade the privacy of users like bypassing encryption and authentication or excessive collection of user data.

bastiaanterhorst:
Thanks for the thoughtful comments @felixbuenemann .

I am not aware of any cases where a license prevented the unethical use of software. I see the hippocratic license as very much on the forefront of a current movement that is asking attention for what is essentially shortcomings of the law. 

Your example of guns is fitting: while you can question how useful it is to sell a gun with a warning label telling people not to use it for harm (and put a legal framework backing that statement in place), it would be even more harmful (and bizarre) to consider a situation where guns were sold _without_ such a notice, and _without_ a legal framework to punish people who do harm others.

That is exactly what we're doing with software.

The Open Source movement has frequently spoken in favor of complete freedom, as opposed to freedom limited by ethics (which is what we practice in the _real world_). Also see [here](https://opensource.org/faq#evil). I believe that to be naive, and untenable in today's world. 

Is the hippocratic license going to fix all of this? Nope. But I believe it is a step in the right direction, and a much needed conversation starter.

felixbuenemann:
Here's a proposed diff against the original LICENSE.txt that avoids the whitespace changes:

```diff
--- LICENSE.txt	2019-11-01 11:22:34.000000000 +0100
+++ LICENSE.txt	2019-11-01 11:22:18.000000000 +0100
@@ -8,8 +8,28 @@
 permit persons to whom the Software is furnished to do so, subject to
 the following conditions:
 
-The above copyright notice and this permission notice shall be
-included in all copies or substantial portions of the Software.
+* The above copyright notice and this permission notice shall be
+  included in all copies or substantial portions of the Software.
+
+* No Harm: The software may not be used by anyone for systems or
+  activities that actively and knowingly endanger, harm, or otherwise
+  threaten the physical, mental, economic, or general well-being of
+  other individuals or groups, in violation of the United Nations
+  Universal Declaration of Human Rights
+  (https://www.un.org/en/universal-declaration-human-rights/).
+
+* Services: If the Software is used to provide a service to others, the
+  licensee shall, as a condition of use, require those others not to use
+  the service in any way that violates the No Harm clause above.
+
+* Enforceability: If any portion or provision of this License shall to
+  any extent be declared illegal or unenforceable by a court of
+  competent jurisdiction, then the remainder of this License, or the
+  application of such portion or provision in circumstances other than
+  those as to which it is so declared illegal or unenforceable, shall
+  not be affected thereby, and each portion and provision of this
+  Agreement shall be valid and enforceable to the fullest extent
+  permitted by law.
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
@@ -18,3 +38,7 @@
 LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+This Hippocratic License is an Ethical Source license
+(https://ethicalsource.dev) derived from the MIT License, amended to
+limit the impact of the unethical use of open source software.
```

julik:
@felixbuenemann Thanks for your input! I will apply the diff shortly, having it formatted in a way that reveals the changes better is definitely a good idea.

In terms of influence on the downstream dependents: yes, I do believe the license is going to be "viral", that's why I am intending to support the 4.x branch in so far as required for the dependents who are not able to accept the conditions of this change - for whatever reason. At the very least bug fixes, and probably perf optimisations, are still going to get backported. 

> That said I'm skeptical of the real world impact of such a license.

To an extent I am too, but I do believe that we as a community should start somewhere. So far even enforcing the GPL has proven a challenge for some projects.

Of course a government or a nation state can, and will, use anything they want in any way they see fit and are unlikely to get reprimanded for it. However, a private consultancy or company which furnishes services to such entities is going to be liable because they usually know who they are delivering to. I believe the primary motion of the license is making actors such as consultancies way less comfortable providing the abusive entities with software on behalf of the software's original authors. And it seems that in that role the license does have potential.

felixbuenemann:
@julik Thanks for your feedback.

I think coupling the license change with a major version bump is a good idea and I will follow suit in xlsxtream by releasing a new major version that allows the usage with zip_tricks 5.x when this is released.

fringd:
Hmmm the license seems well intentioned, but it seems so hard to enforce. If I license my code under MIT it's not very meaningful and free if people can't really use it because I depend on code with a more restrictive license. A lot of companies who do no do harm would still avoid this because they don't want to be in court arguing about whether or not they do harm.

If you guys are changing license for sure though, I think having a major version bump would help for people that want to avoid the new license.

julik:
> If you guys are changing license for sure though, I think having a major version bump would help for people that want to avoid the new license.

@fringd Thanks for your input. Yes, this is indeed what we are doing. We are not going to "delicense" zip_tricks 4.x from zipline and for zipline it might be safer to stick to the 4.x version specifier if you are not comfortable trying the hippocratic license on it.

felixbuenemann:
FYI xlsxtream >= 2.2 supports zip_tricks >= 4.5 and 5.x.

I was a bit stressed when I did the release, so I forgot I wanted to do a major version bump and just followed semver, which made this a feature release…

mvz:
Has there been some independent review/article about the consequences of this license? I have a lot of questions and the FAQ doesn't answer them or isn't very clear in its answers.

mvz:
I still have some questions:
* Can downstream dependencies stay MIT-licensed if they depend on zip_tricks 5.x?
* Will you be doing security updates on the 4.x version?

PR #63, opened 2020-01-28
Update syntax and wetransfer_style for Ruby 2.7

julik:
The code also works "as-is" now on 2.7 with few warnings, but it is nicer to be up-to-date. We do have to wait for rspec-mocks to be updated though.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,8 @@ Layout/FirstMethodArgumentLineBreak:
   Enabled: false
 Layout/FirstMethodParameterLineBreak:
   Enabled: false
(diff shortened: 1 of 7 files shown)


julik:
Closing because:

* Changes for syntax are cosmetic and mostly are related to the updated Rubocop rules
* We cannot use Rubocop attached to wetransfer_style since it requires Ruby 2.3+, while zip_tricks is compatible with 2.1+. There is no reason to remove 2.1 compatibility just to satisfy a set of syntax rules which are set by ourselves
* The only warnings that are likely to be experienced on 2.7 are ones from tests
* rspec-mocks hasn't been updated yet

PR #64, opened 2020-01-28
Improve RemoteUncap

julik:
Working with S3 it is better to determine the object size using a GET request for one byte.

The API surface of RemoteUncap is also made smaller. To test RemoteUncap we are much better off using an actual server instead of mocks - so that we are certain everything works. It only takes about half a second to spin up a Puma for that.


.travis.yml
```diff
@@ -1,10 +1,6 @@
 rvm:
-- 2.1
-- 2.2.10
-- 2.3.7
-- 2.4.4
-- 2.5.4
-- 2.6.2
+- 2.1.0
+- 2.6.5
```
linkyndy: We don't want to be sure all supported versions are fine? 🤔

lib/zip_tricks/remote_io.rb
```diff
@@ -9,10 +9,10 @@
 # available.
 class ZipTricks::RemoteIO
   # @param fetcher[#request_object_size, #request_range] an object that perform fetches
```
linkyndy: Please update this one as well.

lib/zip_tricks/remote_io.rb
```diff
@@ -66,12 +65,41 @@ def tell
...
+      content_range_header_value = response['Content-Range']
+      content_range_header_value.split('/').last.to_i
+    when "200"
+      response['Content-Length'].to_i
+    else
+      raise "Remote at #{@uri} replied with code #{response.code}"
+    end
   end
 
   # Reads N bytes at offset from remote
```
linkyndy: Do we still need this method?

lib/zip_tricks/remote_io.rb
```diff
@@ -66,12 +65,41 @@ def tell
...
+  # @return [Integer] the size of the remote resource, parsed either from Content-Length or Content-Range header
   def request_object_size
-    @fetcher.request_object_size
+    http = Net::HTTP.start(@uri.hostname, @uri.port)
+    request = Net::HTTP::Get.new(@uri)
+    request.range = 0..0
+    response = http.request(request)
+    case response.code
+    when "206"
+      content_range_header_value = response['Content-Range']
```
linkyndy: Are these headers always returned with the response? 🤔

spec/zip_tricks/remote_io_spec.rb
```diff
@@ -1,26 +1,18 @@
...
-      expect(subject.read(3)).to eq('abc')
-    end
-  end
+  let(:subject) { described_class.new('http://host.example/some.bin') }
 
   context 'when it internally addresses a remote resource' do
     it 'requests the size of the resource once via #request_object_size and \
         does neet to read if resource is empty' do
-      subject = described_class.new
+      subject = described_class.new('http://host.example/some.bin')
```
linkyndy: If you define `subject` above, you can remove this one.

lib/zip_tricks/remote_io.rb
julik: If the status code is 206 then this header must be present, yes.

.travis.yml
julik: No, I don't think so - the bracket is sufficient (we have minimum compatibility and the newest). If something in-between pops up I rather address that separately.

lib/zip_tricks/remote_io.rb
julik: Good call - it seems to be unused now.

julik:
> Not a big fan of manually starting test servers inside tests, but if you think that's more effective, I trust you.

Hear, but it is very, very effective (we do it in both the download server and nu_backend tests, to great effect).

PR #67, opened 2020-05-04
Fix extended timestamp extra field output

### LINKED ISSUE #65: Tests fail on Rubyzip 2.1
It seems that rubyzip 2.1 changed something and now quite a few tests fail with the following:

```
  8) ZipTricks::OutputEnumerator returns parts of the ZIP file when called via #each with immediate yield
     Failure/Error:
       Zip::File.open(output_buf.path) do |zip_file|
         # Handle entries one by one
         zip_file.each do |entry|
           # The entry name gets returned with a binary encoding, we have to force it back.
           per_filename[entry.name] = entry.get_input_stream.read
         end
       end
     
     NoMethodError:
       undefined method `dos_equals' for nil:NilClass
     # ./spec/zip_tricks/output_enumerator_spec.rb:24:in `block (2 levels) in <top (required)>'

Finished in 26.77 seconds (files took 0.44473 seconds to load)
117 examples, 8 failures

Failed examples:

rspec ./spec/zip_tricks/rack_body_spec.rb:4 # ZipTricks::RackBody is usable as a Rack response body, supports each() and close()
rspec ./spec/zip_tricks/streamer_spec.rb:91 # ZipTricks::Streamer can write and then read the block-deflated files
```

We can either lock rubyzip to a specific version (probably 1.x) or investigate.

julik:
Upon closer inspection of why Rubyzip 2.1 cannot work with our output it turns out that we write the timestamp extra fields incorrectly. To indicate that only mtime is present, the last bit of the flag byte needs to be set. We were erroneously setting the highest bit instead.

Since Rubyzip 2 attempts to read the extra timestamp field, and does it - it doesn't find the flag bit which instructs it to read the mtime from the extra field. Rubyzip _does not_ parse the "standard" time field if the timestamp extra field is present at all - so both the standard time for the entry and the time from the extra field end up being `nil`.

Closes #65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 5.1.1
+
+* Fix extended timestamp extra field output. The first bit of the flag would be set instead of the last bit of
+  the flag, which made it impossible for Rubyzip to read the timestamp of the entry - and it would also make
+  the extra field useless for most reading applications.
+
 ## 5.1.0
 
 * Slightly rework `RemoteIO` and `RemoteUncap` and make sure they work correctly by spinning up a test webserver
--- a/lib/zip_tricks/version.rb
+++ b/lib/zip_tricks/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ZipTricks
-  VERSION = '5.1.0'
+  VERSION = '5.1.1'
 end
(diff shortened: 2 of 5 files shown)


dennisvdvliet:
Is it worth/possible to test this against other versions of rubyzip as well?

Maybe testing against 1.3.0 would be nice since we can expect some people to be on that still.

julik:
It is possible but I don't think it is worth it. If we do, our test matrix grows 2x because we will then need separate Gemfiles for those tests, and there is also differentiation on the Ruby version (Rubyzip 2.x won't work on the older Rubies we want to stay compat with). Since the bug was not in Rubyzip but in our code I decided not to do it this time.

julik:
Also good to keep in mind: rubyzip is a development dependency, so if you do not work on zip_tricks _itself_ you won't need to install it or bracket it as a dependency for your application.

PR #70, opened 2020-05-19
Raise on Streamer offset mismatch

julik:
To prevent a situation where a user might forget to use `sumulate_write` -
we want to make this failure mode explicit and early. Otherwise an invalid
ZIP might get produced.

This verification is performed when the central directory gets written -
we could fail earlier with it, but then we would need to walk all written
entries in order, after each added entry. This would make the algorithm
accidentally quadratic. If we want better performance we need to add an
additional offset tracking, and update it on every entry. This will allow
us to fail earlier, but is a bit of a drag to implement correctly.

To me it seems that failing at central directory output is a sensible compromise.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,3 +11,5 @@ Style/GlobalVars:
       - qa/*.rb
       - spec/spec_helper.rb
       - spec/support/zip_inspection.rb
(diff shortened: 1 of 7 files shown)


julik:
Yes, we should! Forgot about it

PR #72, opened 2020-06-27
Fix Ruby 2.7 warning due to bad indentation

felixbuenemann:
This fixes the following Ruby 2.7+ warning:

> lib/zip_tricks/zip_writer.rb:198: warning: mismatched indentations at 'end' with 'else' at 196


felixbuenemann:
Since this is a whitespace only change I went ahead and merged this without review from other contributors.

julik:
🚀

PR #74, opened 2020-09-13
Get rid of Rails streaming in favor of iterables

julik:
The Rails Live responses do not work correctly with
RSpec and other integration tests, and are fiddly.

It seems that the Zipline setup works better.


felixbuenemann:
@julik Fixed the broken spec for this one.

felixbuenemann:
@julik LGTM, anything left to do?

It's great that dependency of live streaming is dropped.

PR #75, opened 2020-09-15
Fix FileReader example code

hibariya:
Hello. When I tried to use this for testing, I noticed the example code did not work and fixed it.


julik:
Thank you 🙏

Issue #78, opened 2020-11-05
Using RackBody does many many many write() syscalls


WJWH:
The `RackBody` class spins up a `Streamer` which in turn spins up a `ZipWriter`, passing along an io-like object. The `ZipWriter` will `<<` many extremely small strings into this io object, often only one or two bytes for local header values. Puma handles this rather poorly, since it essentially does `body.each {|chunk| socket.syswrite(chunk)}`. This leads to dozens/hundreds of very small `write()` syscalls and all the context switches to the kernel generates quite a bit of overhead. This can be verified by starting the script below with `strace -f puma syscall_ziptricks_example.ru` and observing the syscalls generated.

Reproduction script:
```ruby
# syscall_ziptricks_example.ru
require 'zip_tricks'

class ChonkyBoi
  CHONK_SIZE = 16384 # default-ish linux socket write buffer size 

  def initialize(body)
    @body = body
  end

  def each
    buffer = String.new(capacity: CHONK_SIZE).force_encoding(Encoding::BINARY)
    current_buffer_size = 0
    @body.each do |chonk|
      chonk_size = chonk.bytesize
      if current_buffer_size + chonk.bytesize > CHONK_SIZE
        yield buffer
        buffer = String.new(capacity: CHONK_SIZE).force_encoding(Encoding::BINARY)
        current_buffer_size = 0
      end
      buffer << chonk
      current_buffer_size += chonk_size
    end
    yield buffer
  end
end

app = ->(env) {
  body = ZipTricks::RackBody.new do | zip |
    zip.write_stored_file('syscall_ziptricks_example.ru') do |sink|
      File.open('broken_ziptricks_example.ru', 'rb'){|source| IO.copy_stream(source, sink) }
    end
  end
  # [200, {}, ChonkyBoi.new(body)]
  [200, {}, body]
}

run app
```

The Run the script with `puma syscall_ziptricks_example.ru` and then test using `wrk` with `wrk -c 2 -d 30`. Using the non-chunked body, I get about 600-650 requests per second. The most straightforward way to reduce syscall overhead is to do more work per syscall, so accumulating the smaller strings into a buffer first has a big effect. Changing the example to use the chunked example, throughput goes up to about 2600-2750 req/s.

julik:
[...]

PR #79, opened 2020-11-05
Buffer metadata writes in the Streamer

julik:
POC looking into #78

### DIFF
--- a/bench/buffered_file_writes_bench.rb
+++ b/bench/buffered_file_writes_bench.rb
@@ -1,0 +1,63 @@
+require 'bundler'
+Bundler.setup
+
+require 'benchmark'
+require 'benchmark/ips'
+require_relative '../lib/zip_tricks'
+
+buffer_sizes = (0..128).map { |n| n * 1024 }
+
+require 'benchmark/ips'
+require 'tempfile'
+
+rng = Random.new
+# For a more accurate simulation let's assume that the
+# files are being written into the ZIP piecewise, in chunks of
+# 16KB.
+blobs = (1..64).map { rng.bytes(16 * 1024) }
+n_files = 256
+
+Benchmark.ips do |x|
+  x.config(time: 5, warmup: 0)
+  buffer_sizes.each do |buf_size|
+    x.report "Writes using a #{buf_size} byte buffer" do
+      tf = Tempfile.new
+      ZipTricks::Streamer.open(tf, write_buffer_size: buf_size) do |zip|
+        n_files.times do |n|
+          zip.write_stored_file("file-#{n}") do |sink|
+            blobs.each do |blob_str|
+              sink << blob_str
+            end
+          end
+        end
+      end
+    ensure
+      tf.close
+    end
+  end
+  x.compare!
+end
+
+__END__
+
+Calculating -------------------------------------
+Writes using a 0 byte buffer
+                          2.458  (± 0.0%) i/s -     12.000  in   5.289171s
+Writes using a 512 byte buffer
+                          2.584  (± 0.0%) i/s -     13.000  in   5.046729s
+Writes using a 1024 byte buffer
+                          2.466  (± 0.0%) i/s -     13.000  in   5.307981s
+Writes using a 16384 byte buffer
+                          2.373  (±42.1%) i/s -     10.000  in   5.026637s
+Writes using a 32768 byte buffer
+                          2.572  (± 0.0%) i/s -     13.000  in   5.088180s
+Writes using a 65536 byte buffer
+                          2.594  (± 0.0%) i/s -     13.000  in   5.031270s
+
+Comparison:
+Writes using a 65536 byte buffer:        2.6 i/s
+Writes using a 512 byte buffer:        2.6 i/s - 1.00x  slower
+Writes using a 32768 byte buffer:        2.6 i/s - 1.01x  slower
+Writes using a 1024 byte buffer:        2.5 i/s - 1.05x  slower
+Writes using a 0 byte buffer:        2.5 i/s - 1.06x  slower
+Writes using a 16384 byte buffer:        2.4 i/s - same-ish: difference falls within error
--- a/bench/buffered_file_writes_bench.rb
+++ b/bench/buffered_file_writes_bench.rb
@@ -5,38 +5,33 @@ require 'benchmark'
 require 'benchmark/ips'
 require_relative '../lib/zip_tricks'
 
-buffer_sizes = (0..128).map { |n| n * 1024 }
+buffer_sizes = (0..4).map { |n| n * 1024 }
 
 require 'benchmark/ips'
 require 'tempfile'
 
 rng = Random.new
-# For a more accurate simulation let's assume that the
-# files are being written into the ZIP piecewise, in chunks of
-# 16KB.
-blobs = (1..64).map { rng.bytes(16 * 1024) }
-n_files = 256
+n_files = 1024
+tf = Tempfile.new
+blob = rng.bytes(1024)
 
 Benchmark.ips do |x|
   x.config(time: 5, warmup: 0)
   buffer_sizes.each do |buf_size|
     x.report "Writes using a #{buf_size} byte buffer" do
-      tf = Tempfile.new
       ZipTricks::Streamer.open(tf, write_buffer_size: buf_size) do |zip|
         n_files.times do |n|
           zip.write_stored_file("file-#{n}") do |sink|
-            blobs.each do |blob_str|
-              sink << blob_str
-            end
+            8.times { sink << blob }
           end
         end
       end
-    ensure
-      tf.close
     end
   end
   x.compare!
 end
+
+tf.close
 
 __END__
 
--- a/README.md
+++ b/README.md
@@ -127,11 +127,14 @@ ZipTricks::Streamer.open(io) do | zip |
   # Write the local file header first..
   zip.add_stored_entry(filename: "first-file.bin", size: raw_file.size, crc32: raw_file_crc32)
 
-  # then send the actual file contents bypassing the Streamer interface
+  # Adjust the ZIP offsets within the Streamer. It is important to do this first
+  # so that any buffered data gets flushed from the write buffer and into the IO, before
+  # the "bypassing" writes take place
+  zip.simulate_write(my_temp_file.size)
+
+  # ...and then send the actual file contents bypassing the Streamer interface
   io.sendfile(my_temp_file)
 
-  # ...and then adjust the ZIP offsets within the Streamer
-  zip.simulate_write(my_temp_file.size)
 end
 ```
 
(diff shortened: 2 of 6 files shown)


julik:
Right - so I've done some work here but it does introduce a bit of an intricacy in that the "boundaries" of the flushed buffered pages may end up "between" the local file headers and the like. Now, _normally_ this should not create any issues except when using bypassing writes _and_ doing it in the order recommended in the documentation, blegh. If we go for this approach then we need to requier that `simulate_write` gets called _first._ And there is no way to implement an exception for this because we can't know whether a bypass write has just been performed when `simulate_write` gets called.

An alternative would be to always flush the buffer at "meanigful object" boundary - so after every local file header, data descriptor and so on 🤔

julik:
@WJWH @felixbuenemann do you have ideas how I could narrow down on the bufsize that is right? With a local file the numbers are real close to each other except for when no buffering is used - then it's slower

felixbuenemann:
@julik I'm still looking into this and reading zlib codebase to get a better understanding.

Internally zlib buffer 64K for CRC32 when using deflate with hashing (not currently used, due to -MAX_WBITS). For pending writes there is a 64K (DEF_MEM_LEVEL) or 128K (MAX_MEM_LEVEL) buffer, so max 64K compressed data will currently be buffered.

For writing data into deflate the LZ77 compression uses a window size of 32K, so any data written can reference sequences from the previous 32K.

If you have non-compressible blocks, they are limite to 64K.

So I woulkd say 64K is a safe default as a write buffer.

Since we're not flushing for each write compression efficiency is not affected by block size.

I also noticed that the `ZipTricks::Streamer::DeflatedWriter` flushes every 5 MB of input data, which I think is redundant, since in my testing Zlib already flushed every 4,226,305 Bytes or `(4<<20)+(31<<10)+(1<<8)+1`, when writing all zeroes to deflate and using `Zlib::DEF_MEM_LEVEL`. When using `Zlib::MAX_MEM_LEVEL` this roughly doubles to 8,453,377 Bytes or `(8<<20)+(63<<10)+(1<<8)+1`.

I also noticed that we can implement `ZipTricks::Streamer::DeflatedWriter` only using deflate, without extra streaming CRC32, by using Gzip (MAX_WBITS+16) instead of Raw (-MAX_WBITS) format, skipping the Gzip header and cutting of the Gzip trailer to get CRC32 and uncompressed size. This removes most of the code from DeflatedWriter, but is neither slower nor faster. I haven't looked at memory usage though.

Here's the modified version of `ZipTricks::Streamer::DeflatedWriter` without doc comments:

```ruby
class ZipTricks::Streamer::DeflatedWriter
  def initialize(io)
    @compressed_io = ZipTricks::WriteAndTell.new(io)
    @deflater = ::Zlib::Deflate.new(::Zlib::DEFAULT_COMPRESSION, ::Zlib::MAX_WBITS+16)
    @deflater.flush # discard gzip header
  end

  def <<(data)
    @compressed_io << @deflater.deflate(data)

    self
  end

  def finish
    until @deflater.finished?
      data = @deflater.finish
      # remove gzip trailer and extract crc32 and uncompressed size
      crc32, isize = data.slice!(-8, 8).unpack("VV") if @deflater.finished?
      @compressed_io << data # write remaining data
    end
    {crc32: crc32, compressed_size: @compressed_io.tell, uncompressed_size: isize}
  end
end
```

julik:
Interesting. Although I think the issue that Wander was having was primarily that every "bit" of data we write about a ZIP entry caused a very tiny write() on the Puma side. Buffering while deflating should already be covered with the current settings. And I'm not really sold on the "clever" crc32 extraction approach because, for one, the `V` specifier means that the uncompressed size of a single file is limited to a uint32. I think zip64 allows larger files.

What if we settle on only batching writes of the metadata segments?

felixbuenemann:
Sorry, I wrote a ￼reply to this, but closed the browser before sending it. 

Yeah, the gzip hack is mood due to zip64, didn't think about that. 

If you want to buffer metadata writes the easiest way is probably to buffer all writes, if the destination is again something IO like. 

I haven't looked at your changes yet though.

bench/buffered_file_writes_bench.rb
```diff
@@ -0,0 +1,58 @@
+require 'bundler'
+Bundler.setup
+
+require 'benchmark'
+require 'benchmark/ips'
+require_relative '../lib/zip_tricks'
+
+buffer_sizes = (0..4).map { |n| n * 1024 }
```
felixbuenemann, line 8: The Benchmark fails to test the slowest range, which is below 512 bytes…

felixbuenemann: One thing to keep in mind is then when I/O is coming from a block device than the kernel will  handle anything below the block size (usually 512 Byte) by reading the same 512 Byte block over and over again and discarding the bytes that aren't needed.

You can try this with `dd` on Linux:

```shell
$ dd if=/dev/zero of=in.dat bs=1M count=32
128+0 records in
128+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.922853 s, 145 MB/s
# go make a coffee before you run this:
dd if=in.dat of=out.dat bs=1
$ dd if=in.dat of=out.dat bs=1 iflag=direct oflag=direct
dd: error reading 'in.dat': Invalid argument
0+0 records in
0+0 records out
0 bytes copied, 0.000204668 s, 0.0 kB/s
$ dd if=in.dat of=out.dat bs=512
262144+0 records in
262144+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 1.79363 s, 74.8 MB/s
# Direct I/O works from bs=512, but much slower, due to lack of buffered I/O optimizations like readahead
$ dd if=in.dat of=out.dat bs=512 iflag=direct oflag=direct iflag=direct oflag=direct
262144+0 records in
262144+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 45.6457 s, 2.9 MB/s
```

Doing a full block size sweep shows throughput leveling of after about 4K (with buffered I/O that is):

```
+ for bsize in 1 2 4 8 16 32 64 128 256 512 1K 2K 4K 8K 16K 32K 64K 128K 256K 512K 1M 2M
+ dd if=in.dat of=out.dat bs=1
134217728+0 records in
134217728+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 395.099 s, 340 kB/s
+ dd if=in.dat of=out.dat bs=2
67108864+0 records in
67108864+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 200.144 s, 671 kB/s
+ dd if=in.dat of=out.dat bs=4
33554432+0 records in
33554432+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 99.8447 s, 1.3 MB/s
+ dd if=in.dat of=out.dat bs=8
16777216+0 records in
16777216+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 51.6151 s, 2.6 MB/s
+ dd if=in.dat of=out.dat bs=16
8388608+0 records in
8388608+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 24.0579 s, 5.6 MB/s
+ dd if=in.dat of=out.dat bs=32
4194304+0 records in
4194304+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 12.6482 s, 10.6 MB/s
+ dd if=in.dat of=out.dat bs=64
2097152+0 records in
2097152+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 6.34185 s, 21.2 MB/s
+ dd if=in.dat of=out.dat bs=128
1048576+0 records in
1048576+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 3.8795 s, 34.6 MB/s
+ dd if=in.dat of=out.dat bs=256
524288+0 records in
524288+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 2.4245 s, 55.4 MB/s
+ dd if=in.dat of=out.dat bs=512
262144+0 records in
262144+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 1.57312 s, 85.3 MB/s
+ dd if=in.dat of=out.dat bs=1K
131072+0 records in
131072+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 1.34376 s, 99.9 MB/s
+ dd if=in.dat of=out.dat bs=2K
65536+0 records in
65536+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.957081 s, 140 MB/s
+ dd if=in.dat of=out.dat bs=4K
32768+0 records in
32768+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.831601 s, 161 MB/s
+ dd if=in.dat of=out.dat bs=8K
16384+0 records in
16384+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.76059 s, 176 MB/s
+ dd if=in.dat of=out.dat bs=16K
8192+0 records in
8192+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.836844 s, 160 MB/s
+ dd if=in.dat of=out.dat bs=32K
4096+0 records in
4096+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.750531 s, 179 MB/s
+ dd if=in.dat of=out.dat bs=64K
2048+0 records in
2048+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.724429 s, 185 MB/s
+ dd if=in.dat of=out.dat bs=128K
1024+0 records in
1024+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.774725 s, 173 MB/s
+ dd if=in.dat of=out.dat bs=256K
512+0 records in
512+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.891007 s, 151 MB/s
+ dd if=in.dat of=out.dat bs=512K
256+0 records in
256+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.852597 s, 157 MB/s
+ dd if=in.dat of=out.dat bs=1M
128+0 records in
128+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.758093 s, 177 MB/s
+ dd if=in.dat of=out.dat bs=2M
64+0 records in
64+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 0.724734 s, 185 MB/s
```

Since the kernel uses a page size of 4K it maskes sense, that this is the sweet spot to target, unless there are special circumstances to leverage a larger block size, eg. spp code efficiencies or other internal buffers, like the 64K CRC buffer in Zlib.

felixbuenemann: Oh and of course the use case matters. For example I'm working on [plexdrive](https://github.com/plexdrive/plexdrive), which is a FUSE filesystem written in Go.

In plexdrive I got a *huge* efficiency improvement by going from 4K to 128K I/Os by changing the readahead size. Hoewever you're in quite different performance land when you're implementing a filessystem that can do several hundred megabytes per second throughput.

julik:
Closing in favor of https://github.com/WeTransfer/zip_tricks/pull/84

PR #80, opened 2020-11-05
Speed Up CRC32 Calculation

felixbuenemann:
Avoid reinitializing CRC32 state for each buffer write. Gives about 40x speedup for smaller writes.


julik (approved):
😍

PR #81, opened 2020-11-05
Remove QA files and scripts

julik:
The test suite of ZipTricks itself is mature enough to capture most
of the edge cases, and in the last few issues that we had we didn't have
situations where something was caught testing using the QA data and wasn't
caught in the unit tests.

Since end-to-end tests call to zipinfo and BOMArchiveHelper anyway, having
more things bloat the repo is probably not a very good idea anymore. Also, since
those QA things do not run as part of the CI pipeline, there is little incentive
to run them regularly.


felixbuenemann (approved):
No objections.

PR #83, opened 2020-11-06
Fix hung specs running Archive Utility on macOS

felixbuenemann:
* Skip unless we're running macOS (Darwin)
* Use `open` to force separate copy and wait for exit
* Run using bundle id `com.apple.archiveutility` instead of path
* Temporarily change dearchive destination to cwd
* Temporarily disable showing folder in Finder after uncompress
* Check folder contents against provided list of entries

### DIFF
--- a/spec/support/zip_inspection.rb
+++ b/spec/support/zip_inspection.rb
@@ -1,3 +1,5 @@
+require 'rbconfig'
+
 module ZipInspection
   def inspect_zip_with_external_tool(path_to_zip)
     zipinfo_path = 'zipinfo'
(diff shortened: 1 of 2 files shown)


spec/support/zip_inspection.rb
```diff
@@ -11,19 +13,35 @@ def inspect_zip_with_external_tool(path_to_zip)
...
+    unless RbConfig::CONFIG['host_os'].match? /darwin|mac os/
+      skip "Archive Utility specs require macOS"
+      return
+    end
+    # ArchiveUtility sometimes puts the stuff it unarchives in ~/Downloads etc. so
+    # we reset the preferences to unpack to current dir and don't show in Finder.
+    dearchive_reveal_after = `/usr/bin/defaults read com.apple.archiveutility dearchive-reveal-after`.to_i == 1
+    dearchive_into = `/usr/bin/defaults read com.apple.archiveutility dearchive-into`.chomp
+    begin
+      system *%W[/usr/bin/defaults write com.apple.archiveutility dearchive-reveal-after -bool false]
```
julik, line 34: I'm sorry but I think we shouldn't be doing this. Changing preferences for end-user software (even if we do it in this safe manner, and because the said software is so obtuse that it cannot be configured in a different manner) is not something a test suite for a library should be doing.

Also I actually did look into the folders that archive utility was producing - it is manual but it was working more or less OK. A case could be made for removing this test altogether though...

spec/support/zip_inspection.rb
```diff
@@ -11,19 +13,35 @@ def inspect_zip_with_external_tool(path_to_zip)
...
 
-  def open_zip_with_archive_utility(path_to_zip, skip_if_missing: false)
-    # ArchiveUtility sometimes puts the stuff it unarchives in ~/Downloads etc. so do
-    # not perform any checks on the files since we do not really know where they are on disk.
-    # Visual inspection should show whether the unarchiving is handled correctly.
-    au_path = '/System/Library/CoreServices/Applications/Archive Utility.app/' \
-              'Contents/MacOS/Archive Utility'
-    open_with_external_app(au_path, path_to_zip, skip_if_missing)
+  def open_zip_with_archive_utility(path_to_zip, skip_if_missing: false, expected_content: [])
+    unless RbConfig::CONFIG['host_os'].match? /darwin|mac os/
```
julik, line 25: Can't this be done with `RUBY_PLATFORM`?

julik (changes requested):
Not quite ok with the `defaults write` manoever. Also - where was this command hanging? at the archive utility invocation itself? And on which OS version (I am a bit of a caveman and still have Mojave on all my boxes)

spec/support/zip_inspection.rb
felixbuenemann: Yes, except on JRuby:

```
# uname -msr
Darwin 19.6.0 x86_64
# rvm ruby-2.7 exec irb
2.7.2 :001 > RUBY_PLATFORM
 => "x86_64-darwin19"
2.7.2 :002 > RbConfig::CONFIG['host_os']
 => "darwin19.6.0"
# rvm jruby exec irb
jruby-9.2.9.0 :001 > RUBY_PLATFORM
 => "java"
jruby-9.2.9.0 :002 > RbConfig::CONFIG['host_os']
 => "darwin"
# rvm truffleruby exec irb
truffleruby-20.2.0 :001 > RUBY_PLATFORM
 => "x86_64-darwin18"
truffleruby-20.2.0 :002 > RbConfig::CONFIG['host_os']
 => "darwin"
```

Interestingly truffleruby returns a wrong value for `RUBY_PLATFORM`, since it is running on Darwin 19 not 18.

I'm not sure why the "mac os" value is in there, should be "darwin" everywhere.

spec/support/zip_inspection.rb
felixbuenemann: I got annoyed with the Finder window constantly popping up and the temp folder is removed at the end of the specs, so I'm curious how syou managed to inspect the folders in the split second it exists.

The folder contents are now checked using ruby during the specs, so no need for manual inspection. We could at file size or CRC checks easily as well.

Since Archive Utility is a shitty tool to automate I wouldn't mind removing it alltogether or make the tests run only manually using an rspec tag.

Unless we want automated tests to unarchive files in some unknown destination on the developers Mac, there's no way around changing the defaults, unless we just check the values and crash if configured badly.

Given that, changing the defaults and setting them back with ensure seems a good compromise.

felixbuenemann:
> Also - where was this command hanging? at the archive utility invocation itself? And on which OS version (I am a bit of a caveman and still have Mojave on all my boxes)

I'm on 10.15.7, but it shouldn't really matter. The problem is, that most macOS GUI apps only start one instance and then stay running. So the first time you start Archive Utility it starts and then just keeps running and never quits unless you manually switch to the app and do Cmd+Q or Ctrl+C in the Terminal.

By using pretty much all of the flags to "open" we tell macOS to launch a new instance of the app in the background, even if one is already running and wait for it to exit.

You also can just pass the filename as an argument to the "Archive Utility" binary, it will be ignored, you have to pass is  using open. That probably worked on older macOS versions and is the reason that the app hangs, even if it is the first instance to be launched.

spec/support/zip_inspection.rb
felixbuenemann: @julik By the way, Archive Utility is using libarchive under the hood (on macOS 10.15.7 libarchive/3.3.2), so another way would be to use a gem like "ffi-libarchive" to dlopen it and uncompress the zip.

That would have the benefit of being able to run on Linux as well.

I just tested it and it works fine even on JRuby.

One drawback is that by default it will use a newer libarchive if it is installed by homebrew and only fall back to the system one if not found.

PR #84, opened 2020-11-09
Add write buffering in a few key places

julik:
Another POC for #78


lib/zip_tricks/output_enumerator.rb
```diff
@@ -1,43 +1,61 @@
...
+#     end
+#
+#     [200, {'Content-Type' => 'binary/octet-stream'}, iterable_zip_body]
 class ZipTricks::OutputEnumerator
-  # Prepares a new Rack response body with a Zip output stream.
-  # The block given to the constructor will be called when the response
-  # body will be read by the webserver, and will receive a {ZipTricks::Streamer}
-  # as it's block argument. You can then add entries to the Streamer as usual.
-  # The archive will be automatically closed at the end of the block.
+  DEFAULT_WRITE_BUFFER_SIZE = 65 * 1024
```
felixbuenemann, line 28: Is the 65 vs 64 a typo or intentional?

felixbuenemann:
@julik Could you auf s high-level overview of the changes and what you are trying to achieve to the PR description?

lib/zip_tricks/output_enumerator.rb
julik: Non intentional

julik:
Yes, sorry - my bad. What I wanted to figure out is what kind of buffering we could use and where to introduce it while not breaking the library too much. I have tried the following approaches:

* Blanket output buffering on the Streamer level - this works, but requires careful management when a sideband write gets performed. There is a non-0 chance of breaking software that uses ZipTricks if we change semantics here (the fact that you have to call `simulate_write` before doing the actual write
* Output buffering in `OutputEnumerator` which works fine except for the case when one does a `#to_a` on the Enumerator or otherwise consumes its output eagerly. Since the `WriteBuffer` reuses one StringIO, which in turn reuses one String, every time the String is written to the underlying output it is the same object. In some situations this will work fine (like in Rack, most likely) but in others it won't (if the output enumerator gets converted to an Array). We might need to dup the String when we write it, which might lead to elevated memory use or increase allocations. This is in the branch called `buffer-at-enum`.
* We can play with the output buffer of the deflater like you have done. However, I think this doesn't pay as much because the deflater already does some buffering for us, it just needs to be configured - or maybe even left alone
* We can introduce write buffering in Streamer but at strictly defined boundaries. This will not be as efficient as using a blanket buffer for the entire Streamer, but it will make sure that the output buffer is flushed at "logical" boundaries of the segments - for example after every local header. If we use that approach we won't need to make the buffer size configurable in Streamer - we can assume that the chunks will be small enough, and we also won't be breaking the consumers of the library, they will still have a guarantee that by the time they need to do something "outside" of the public Streamer methods there is no pending output.
* We can also introduce a buffer between the Writable and the output because otherwise the client performing the writes needs to control the chunking by itself, especially for the StoredWriter which just forwards calls to `<<`. This also seems reasonable to do, but we will have the same tricky thing with the retained string that gets written.

So far it seems like we should go for a combination of buffering in the output enum (you can't really do bypass writes as the write destination is not an IO), buffering at logical segment boundaries of the ZIP and - maybe - buffering in Writable. Then regardless of how the client does chunking we will have consistent output.

And yet another alternative: we buffer only at the OutputEnumerator (which will improve performance for @WJWH and for all users of zipline) and state that buffering is for the consumer of the library to customise in other use cases. This will be cheaper in terms of the amount of tricky measurement we would otherwise have to do. I say it would be tricky because the performance really depends on how many files you have, how big they are and whether they are compressed or stored.

The benchmark can be discarded.

felixbuenemann:
The `OutputEnumerator` case with `#to_a` etc. shouldn't be a problem, just always `#dup` the string, it uses copy-on-write semantics, at least on MRI, so it means there's no new allocation unless the duped string is modified and the other way around.

julik:
Interesting - do you have a link to documentation where CoW is documented for `String#dup`?

felixbuenemann:
> Interesting - do you have a link to documentation where CoW is documented for `String#dup`?

I don't have, but a quick test shows that both `String#dup` and `String#clone` do not increase memory footprint (RSS):

```ruby
s = " " * (10<<20) # MiB
a = Array.new(100)

def current_rss
  `ps -o rss= -p #$$`.to_i/1024.0
end

def test_allocations(label)
  pid = fork do
    h = ObjectSpace.count_objects
    rss_beg = current_rss
    GC.disable
    GC.start(full_mark: true, immediate_sweep: true)
    num_string_beg = ObjectSpace.count_objects(h)[:T_STRING]
    yield
    num_string_end = ObjectSpace.count_objects(h)[:T_STRING]
    rss_end = current_rss
    GC.enable
    puts "%16s - Strings allocated: %5d – Memory allocated: %9.3f MiB" % [
      label, (num_string_end - num_string_beg), (rss_end - rss_beg)
    ]
  end
  Process.wait2 pid
end

puts "Initial RSS: %.3f MiB (includes %d MiB test String)" % [current_rss, s.bytesize>>20]

test_allocations "Nop (Reference)" do
  for i in 0...a.size
    a[i] = s
  end
end

test_allocations "Dup (CoW)" do
  for i in 0...a.size
    a[i] = s.dup
  end
end

test_allocations "Dup (Mutation)"do
  for i in 0...a.size
    a[i] = s.dup << i
  end
end

test_allocations "Clone (CoW)" do
  for i in 0...a.size
    a[i] = s.clone
  end
end

test_allocations "Clone (Mutation)"do
  for i in 0...a.size
    a[i] = s.clone << i
  end
end
```

```shell
$ ruby -v
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]
$ ruby alloc_test.rb
Initial RSS: 23.406 MiB (includes 10 MiB test String)
 Nop (Reference) - Strings allocated:     0 – Memory allocated:     2.434 MiB
       Dup (CoW) - Strings allocated:   101 – Memory allocated:     2.477 MiB
  Dup (Mutation) - Strings allocated:   101 – Memory allocated:  1012.977 MiB
     Clone (CoW) - Strings allocated:   101 – Memory allocated:     2.473 MiB
Clone (Mutation) - Strings allocated:   101 – Memory allocated:  1012.953 MiB
```

lib/zip_tricks/streamer.rb
```diff
@@ -147,11 +147,14 @@ def self.output_enum(**kwargs_for_new, &zip_streamer_block)
   def initialize(stream, writer: create_writer, auto_rename_duplicate_filenames: false)
     raise InvalidOutput, 'The stream must respond to #<<' unless stream.respond_to?(:<<)
 
-    @dedupe_filenames = auto_rename_duplicate_filenames
-    @out = ZipTricks::WriteAndTell.new(stream)
+    # A small buffer used primarily to avoid too many microscopic writes. This helps a lot
+    # with serving out the data via network
+    @buf = ZipTricks::WriteBuffer.new(stream, 4 * 0124)
```
felixbuenemann, line 152: Another typo here 0124 (84) vs 1024.

Btw. you can also use 1<<10 for KiB and 2<<20 for MiB, so 4K would be 4<<10.

lib/zip_tricks/streamer.rb
```diff
@@ -182,6 +185,7 @@ def write(binary_data)
   # @param num_bytes [Integer] how many bytes are going to be written bypassing the Streamer
   # @return [Integer] position in the output stream / ZIP archive
   def simulate_write(num_bytes)
+    @buf.flush!
```
felixbuenemann, line 188: There is a los of manual flushing going on here. Shouldn't `WriteAndTell` do the flusing on its own, if the underlying IO `responds_to?(:flush)`? I would also rename `flush!` to `flush` to be in line with the real IO class.

julik: Good idea, but we'll need to keep the alias available as it is a public method. Re. WriteAndTell doing the flushing - why would that work well? Is it good to place that responsibility with WriteAndTell to begin with?

There is indeed a lot of manual flushing but that's necessary to ensure we flush after every meaningful segment 🤔

felixbuenemann:
@julik I made a similar change to your buffer write passthrough in https://github.com/WeTransfer/zip_tricks/pull/84/commits/a6cc8b5584bf6239068eec9ad9c21a12601bd65b and it conflicted with your changes, so I merged both, but kept my logic.

The difference in my change is that in ensures buffer size is never exceeded, but also ensures writes are always a multiple of the buffer size, at least for the final write or if data is manually flushed.

I think this is better, since you solution will trigger small write, big write, small write, big write etc. if you write below buffer size followed by write above buffer size, which effectively makes the buffer useless.

felixbuenemann:
I think one optimization we can add to my solution is to do a fron flush, if the write is an exact multiple of the buffersize to avoid having to split the data just because one unaligned write occured.

felixbuenemann:
I implemented the buffer flush optimization in https://github.com/WeTransfer/zip_tricks/pull/84/commits/7f6e3c14e912481dd2b4fe00bf7323e601d0d088.

julik:
Splendid! Let's get the tests to pass and I'll squash merge? This will need quite a comment in the commit message 😆

felixbuenemann:
The encoding issue on RUby 2.1.0 is a bit puzzling. Looks like on 2.1.0 even though the StringIO is in binmode, the encoding of the underlying string can still change to something else than binary encoding.

That was probably a bug, since it doesn't happen on Ruby 2.1.10, but I could work around it by creating the StringIO from an existing buffer string with binary encoding, which I'm now also using to preallocate the buffer.

felixbuenemann:
@julik Do you mind if I rebase this on master so we get the JRuby Spec fixes?

julik:
Go ahead! Also we need to replace all the remaining `flush!` with `flush` and leave `flush!` only as an alias on WriteBuffer

felixbuenemann:
I can replace the flush calls.

julik:
@felixbuenemann One last thing: can we add a comment to the body of `WriteBuffer#<<` which explains the logic and the reasoning behind it? I think it can be very beneficial as the code is "microscopic" in focus and very detailed, it feels like the intent of the change might be unclear to future contributors?

julik:
We also need to preserve `WriteBuffer#flush!` as an alias for backwards compatibility (see https://github.com/fringd/zipline/pull/64 where I have already applied it)

felixbuenemann:
@julik Yes, I can do that. I'm also trying to get rid of memory allocations. I was previsouly using `String#slice!`, but switched to `String#byteslice` because we want byte not char offsets, but is creates string copies, so I wan't to see if I can get around that.

felixbuenemann:
@julik I've gone through a lot of iterations trying to fight memory bloat in the `WriteBuffer` and there are only two options:

* Don't split the data if it does not fit the buffer and bypass it for that write
* Allow mutation of the strings written to the buffer's writable io

All specs except those that explicitly check for mutation or use an array as a sink work fine when mutating the buffer strings after they've been written to the writable.

The problem is that as soon as we `.dup` the strings written to the writable the next modification of the buffer will cause them to be copied.

It would be possible to add an option to the `OutputEnumerator` and `WriteBuffer` that prevents mutation and uses `.dup` instead.

julik:
Right. Here's my proposal: we document the fact that the string written by the `WriteBuffer` is retained, for memory savings reasons. If the consumer of the library requires the string to be copied they can do so manually. Since ZT is a low level library this seems like a good compromise – all we should do is state our tradeoffs upfront.

Do we _need_ to make the WriteBuffer adhere to the buffer size setting strictly? The only meaningful use case I can think of is something like the example I have added for S3 and which has been mentioned in https://github.com/WeTransfer/zip_tricks/issues/45 where your parts have to be above or at the buffer size. Do you feel like we also need to enforce the "no greater than buffer size" constraint?

julik:
In other words: do we _need_ to make `WriteBuffer` as complicated as we have made it? 😄  or will just a small degree of buffering for the metadata + a blanket buffer on `OutputEnumerator` suffice?

felixbuenemann:
It is mainly as complicated to fit the s3 example, where exact output chunks are expected.

If that is not required and "best effort" is OK, then the best approach is to only buffer what fits and write anything above that out directly, probably with a check if the overflow is bigger than a full buffer to prevent eg a single two byte write, if there was one byte free in the buffer. 

With reasonably small buffers we shouldn't get major memory bloat in that case even without in place mutation.

For example given the strict buffer size requirement of the s3 example memory bloated from about 60M to 230M if the written strings were duped to prevent mutation. 

In the bypass case the same example writes alternating chunks of about 60 bytes headers, the full 5MB as bypass, then a mall buffered write etc. 

So for the examples use case it would be useless, but for what we originally set out to do it would be fine.

julik:
And if we remove the “write through if larger than bufsize” optimization and revert to the previous, “dumb” setup?

felixbuenemann:
> And if we remove the “write through if larger than bufsize” optimization and revert to the previous, “dumb” setup?
> […](#)
> On 15 Nov 2020, at 03:09, Felix Bünemann ***@***.***> wrote: So for the examples use case it would be useless, but for what we originally set out to do it would be fine.

The S3 example works only if the buffer is flushed at the exact size which also mean memory bloat unless we can do in place mutations.

felixbuenemann:
Btw. you can also mutate the source strings written to zip instead of copying/duping them first, but that's imho a very bad idea…

In my code im copying to a second buffer string to avoid additional allocations when splitting the source string. That's why I have another `StringIO` for the incoming data.

felixbuenemann:
@julik I've made a curious disovery while looking into high memory usage for the specs (currently around 1GB):

Memory usage drops a lots, when buffers are divisable by 3K. I tested some stuff like block deflate spec and if I change the buffer to eg 129K instead of 128K memory usage would be let's say 30M instead of 100M.

I have no idea why 3K, but it holds true for me on both Ruby 2.7.2 and 2.1.0.

Do you have any idea why 3K?

julik:
No idea. Fine with documenting this discovery next to the constant for now.

felixbuenemann:
@julik I've been sidetracked a bit, since I found a couple of bugs and odities in other places, I'll create PRs for those soon.

Is it OK if I finish this up this weekend?

julik:
Sure!

felixbuenemann:
@julik I found a huge flaw in our testing that lead to the current buffering in place.

The benchmarks that were used to find the buffer sizes were bogus, since they didn't test the real world. Instead they simulated how fast it would be, if the application did writes at a certain block size.

However inserting a write buffer to aggregate smaller writes from the application doesn't make things faster, it makes things slower. Any buffer code you write will be slower than the overhead for calling a C function, which is very low in MRI and maybe a bit higher with FFI in JRuby.

So if the app does 1 byte writes, adding a buffer in front of that won't make it faster, since you spent the exact same time buffering those IOs up, that you would passing them through unbuffered, except your buffer code takes up cpu time as well. That CPU time is multiplied, if the app makes smaller IOs, so buffering up 1 byte writes is a lot slower than lets say 64 byte writes.

This is great news, since it means we can delete a lot of code and replace it with one write buffer in front of the final IO.

If you add a small modification to the CRC buffer size benchmark, you will quickly see that I'm right:

```diff
diff --git a/bench/buffered_crc32_bench.rb b/bench/buffered_crc32_bench.rb
index ba2e6f8..d88a16d 100644
--- a/bench/buffered_crc32_bench.rb
+++ b/bench/buffered_crc32_bench.rb
@@ -28,6 +28,11 @@ require 'benchmark/ips'
 
 Benchmark.ips do |x|
   x.config(time: 5, warmup: 2)
+  x.report "Single-byte <<-writes of #{n_bytes} using no buffer" do
+    crc = ZipTricks::StreamCRC32.new
+    bytes.each { |b| crc << b }
+    crc.to_i
+  end
   buffer_sizes.each do |buf_size|
     x.report "Single-byte <<-writes of #{n_bytes} using a #{buf_size} byte buffer" do
       crc = ZipTricks::WriteBuffer.new(ZipTricks::StreamCRC32.new, buf_size)
```

```
Comparison:
Single-byte <<-writes of 5242880 using no buffer:        1.5 i/s
Single-byte <<-writes of 5242880 using a 65536 byte buffer:        0.1 i/s - 10.91x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 262144 byte buffer:        0.1 i/s - 10.91x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 2097152 byte buffer:        0.1 i/s - 10.93x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 16384 byte buffer:        0.1 i/s - 10.93x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 8192 byte buffer:        0.1 i/s - 10.93x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 131072 byte buffer:        0.1 i/s - 10.94x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 32768 byte buffer:        0.1 i/s - 10.94x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1048576 byte buffer:        0.1 i/s - 10.94x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 524288 byte buffer:        0.1 i/s - 10.95x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1024 byte buffer:        0.1 i/s - 11.00x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 512 byte buffer:        0.1 i/s - 11.05x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 256 byte buffer:        0.1 i/s - 11.10x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1 byte buffer:        0.1 i/s - 12.30x  (± 0.00) slower
```

So yeah, ignoring the unbuffered case, 64K looks to be fastest, but it is over 1000% slower than unbuffered writes. 🐌 

Conversely having the app itself do bigger writes would give a huge speedup:

```diff
diff --git a/bench/buffered_crc32_bench.rb b/bench/buffered_crc32_bench.rb
index ba2e6f8..25f6f04 100644
--- a/bench/buffered_crc32_bench.rb
+++ b/bench/buffered_crc32_bench.rb
@@ -28,12 +28,22 @@ require 'benchmark/ips'
 
 Benchmark.ips do |x|
   x.config(time: 5, warmup: 2)
+  x.report "Single-byte <<-writes of #{n_bytes} using no buffer" do
+    crc = ZipTricks::StreamCRC32.new
+    bytes.each { |b| crc << b }
+    crc.to_i
+  end
   buffer_sizes.each do |buf_size|
     x.report "Single-byte <<-writes of #{n_bytes} using a #{buf_size} byte buffer" do
       crc = ZipTricks::WriteBuffer.new(ZipTricks::StreamCRC32.new, buf_size)
       bytes.each { |b| crc << b }
       crc.to_i
     end
+    x.report "#{buf_size}-byte <<-writes of #{n_bytes} using no buffer" do
+      crc = ZipTricks::StreamCRC32.new
+      bytes.each_slice(buf_size) { |slice| crc << slice.join }
+      crc.to_i
+    end
   end
   x.compare!
 end
```

```
Comparison:
16384-byte <<-writes of 5242880 using no buffer:        3.6 i/s
8192-byte <<-writes of 5242880 using no buffer:        3.5 i/s - 1.03x  (± 0.00) slower
131072-byte <<-writes of 5242880 using no buffer:        3.4 i/s - 1.06x  (± 0.00) slower
32768-byte <<-writes of 5242880 using no buffer:        3.4 i/s - 1.07x  (± 0.00) slower
65536-byte <<-writes of 5242880 using no buffer:        3.3 i/s - 1.10x  (± 0.00) slower
524288-byte <<-writes of 5242880 using no buffer:        3.2 i/s - 1.14x  (± 0.00) slower
2097152-byte <<-writes of 5242880 using no buffer:        3.2 i/s - 1.14x  (± 0.00) slower
262144-byte <<-writes of 5242880 using no buffer:        3.2 i/s - 1.15x  (± 0.00) slower
1048576-byte <<-writes of 5242880 using no buffer:        3.1 i/s - 1.15x  (± 0.00) slower
1024-byte <<-writes of 5242880 using no buffer:        3.1 i/s - 1.16x  (± 0.00) slower
512-byte <<-writes of 5242880 using no buffer:        2.5 i/s - 1.43x  (± 0.00) slower
256-byte <<-writes of 5242880 using no buffer:        2.4 i/s - 1.50x  (± 0.00) slower
Single-byte <<-writes of 5242880 using no buffer:        1.6 i/s - 2.20x  (± 0.00) slower
1-byte <<-writes of 5242880 using no buffer:        0.6 i/s - 5.67x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 2097152 byte buffer:        0.2 i/s - 18.56x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1048576 byte buffer:        0.2 i/s - 18.62x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 524288 byte buffer:        0.2 i/s - 18.74x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 32768 byte buffer:        0.2 i/s - 18.74x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 65536 byte buffer:        0.2 i/s - 18.75x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 262144 byte buffer:        0.2 i/s - 18.77x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 16384 byte buffer:        0.2 i/s - 18.88x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 131072 byte buffer:        0.2 i/s - 19.04x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 8192 byte buffer:        0.2 i/s - 19.08x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 256 byte buffer:        0.1 i/s - 24.69x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1024 byte buffer:        0.1 i/s - 25.23x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 512 byte buffer:        0.1 i/s - 25.27x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1 byte buffer:        0.1 i/s - 28.22x  (± 0.00) slower
```

This example is understating the difference, since looping through the array is inefficient in itself.

If you switch to a `StringIO` the difference becomes much larger:

```ruby
Benchmark.ips do |x|
  x.config(time: 5, warmup: 2)
  n_bytes = 5 << 20 # MiB
  bytes = StringIO.new(Random.new.bytes(n_bytes))
  x.report "Single-byte <<-writes of #{n_bytes} using no buffer" do
    crc = ZipTricks::StreamCRC32.new
    bytes.rewind
    bytes.each(nil, 1) { |b| crc << b }
    crc.to_i
  end
  (0..21).each do |bits|
    buf_size = 1 << bits
    x.report "Single-byte <<-writes of #{n_bytes} using a #{buf_size} byte buffer" do
      crc = ZipTricks::WriteBuffer.new(ZipTricks::StreamCRC32.new, buf_size)
      bytes.rewind
      bytes.each(nil, 1) { |b| crc << b }
      crc.to_i
    end
    x.report "#{buf_size}-byte <<-writes of #{n_bytes} using no buffer" do
      crc = ZipTricks::StreamCRC32.new
      bytes.rewind
      bytes.each(nil, buf_size) { |b| crc << b }
      crc.to_i
    end
  end
  x.compare!
end
```
```
Comparison:
1048576-byte <<-writes of 5242880 using no buffer:      228.8 i/s
524288-byte <<-writes of 5242880 using no buffer:      215.9 i/s - same-ish: difference falls within error
131072-byte <<-writes of 5242880 using no buffer:      210.2 i/s - same-ish: difference falls within error
65536-byte <<-writes of 5242880 using no buffer:      208.7 i/s - same-ish: difference falls within error
262144-byte <<-writes of 5242880 using no buffer:      205.9 i/s - same-ish: difference falls within error
4096-byte <<-writes of 5242880 using no buffer:      202.2 i/s - same-ish: difference falls within error
32768-byte <<-writes of 5242880 using no buffer:      199.5 i/s - same-ish: difference falls within error
2048-byte <<-writes of 5242880 using no buffer:      196.1 i/s - same-ish: difference falls within error
8192-byte <<-writes of 5242880 using no buffer:      193.8 i/s - same-ish: difference falls within error
16384-byte <<-writes of 5242880 using no buffer:      171.5 i/s - 1.33x  (± 0.00) slower
1024-byte <<-writes of 5242880 using no buffer:      170.9 i/s - 1.34x  (± 0.00) slower
2097152-byte <<-writes of 5242880 using no buffer:      149.3 i/s - 1.53x  (± 0.00) slower
512-byte <<-writes of 5242880 using no buffer:      136.8 i/s - 1.67x  (± 0.00) slower
256-byte <<-writes of 5242880 using no buffer:       99.0 i/s - 2.31x  (± 0.00) slower
128-byte <<-writes of 5242880 using no buffer:       62.6 i/s - 3.65x  (± 0.00) slower
64-byte <<-writes of 5242880 using no buffer:       35.4 i/s - 6.45x  (± 0.00) slower
32-byte <<-writes of 5242880 using no buffer:       19.2 i/s - 11.92x  (± 0.00) slower
16-byte <<-writes of 5242880 using no buffer:       19.0 i/s - 12.04x  (± 0.00) slower
8-byte <<-writes of 5242880 using no buffer:        9.5 i/s - 24.16x  (± 0.00) slower
4-byte <<-writes of 5242880 using no buffer:        4.8 i/s - 47.21x  (± 0.00) slower
2-byte <<-writes of 5242880 using no buffer:        2.4 i/s - 94.33x  (± 0.00) slower
Single-byte <<-writes of 5242880 using no buffer:        1.2 i/s - 188.46x  (± 0.00) slower
1-byte <<-writes of 5242880 using no buffer:        1.2 i/s - 189.03x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 524288 byte buffer:        0.2 i/s - 1294.56x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 2097152 byte buffer:        0.2 i/s - 1297.79x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 8192 byte buffer:        0.2 i/s - 1300.06x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 65536 byte buffer:        0.2 i/s - 1301.05x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1048576 byte buffer:        0.2 i/s - 1303.32x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 2048 byte buffer:        0.2 i/s - 1306.33x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 4096 byte buffer:        0.2 i/s - 1306.42x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1024 byte buffer:        0.2 i/s - 1307.16x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 131072 byte buffer:        0.2 i/s - 1309.77x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 32768 byte buffer:        0.2 i/s - 1312.24x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 262144 byte buffer:        0.2 i/s - 1312.65x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 512 byte buffer:        0.2 i/s - 1316.31x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 16384 byte buffer:        0.2 i/s - 1317.45x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 16 byte buffer:        0.2 i/s - 1319.80x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 256 byte buffer:        0.2 i/s - 1324.55x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 128 byte buffer:        0.2 i/s - 1325.53x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 32 byte buffer:        0.2 i/s - 1339.17x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 64 byte buffer:        0.2 i/s - 1347.96x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 4 byte buffer:        0.2 i/s - 1363.96x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 8 byte buffer:        0.2 i/s - 1395.49x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 2 byte buffer:        0.2 i/s - 1407.90x  (± 0.00) slower
Single-byte <<-writes of 5242880 using a 1 byte buffer:        0.2 i/s - 1443.50x  (± 0.00) slower
```

This is using the `ZipTricks::WriteBuffer` in master, which is less then optimal.

The following buffer implementation is 5x faster with the same behavior as master:

```ruby
class ZipTricks::WriteBuffer
  def initialize(writable, buffer_size)
    # Preallocate buffer, then clear it.
    @buf = ("\0".b * (buffer_size * 2)).clear
    @buffer_size = buffer_size
    @writable = writable
  end

  def <<(data)
    if data.bytesize >= @buffer_size
      flush
      @writable << data
    else
      @buf << data
      flush if @buf.bytesize >= @buffer_size
    end
    self
  end

  def flush
    if @buf.bytesize > 0
      @writable << @buf
      @buf.clear
    end
    self
  end

  def to_i
    flush
    @writable.to_i
  end
end
```

Note that `@buf = ("\0".b * (buffer_size * 2)).clear` is equivalent to `String.new(capacity: buffer_size * 2)`, but works on all Ruby versions.

julik:
So how do we optimise this correctly? What I understood from the original issue ( #78 ) was this:

* ZT has a pattern of doing a large number of rapid, tiny writes
* This propagates downstream into the webserver and leads to the webserver performing too many syscalls for inconsequentially small bits of data
* Which leads to a degradation in performance

From there, if we follow where the writes come from, we find:
* The metadata writer (ZipWriter itself) does many very small writes indeed. It is done so that the code if the writer is extremely transparent and easy to follow and change, IMO the payoff here is code clarity. The ease with which ZT was ported to both Elixir and later Crystal proves that it was the right design decision
* The stored file writer - which you should be using if you have precompressed files, such as JPEG or MP4 or MP3 or whatnot - may be performing many tiny writes depending on how the consumer of the writer "portions" the calls. This is an important detail - for instance at WeTransfers the vast majority of the files in the system make zero sense to deflate, using the stored mode is just cheaper. We do not serve them through this particular code path and do it a bit differently, but the point still remains - in a multitude of situations using the stored mode will be more efficient than deflating
* The deflated file writer will have some amount of buffering regardless of the input portioning, because zlib has a buffer so that the compression tables can be used and applied sensibly, and you have already applied there.

From there I can see a few applicable strategies:

* Verify that the deflated writer buffers to some degree (not too much, not too little). Check, you've done this in #96 
* Verify that we do some buffering in the stored writer, this is what we are looking at in this PR
* Verify that we do some buffering when we do metadata writes. This is also what we are looking at in this PR
* Try to apply some buffering for our Rack output, which would also apply in Rails. This is also what we are looking at in this PR

A thing that makes it a bit harder to evaluate things is that writing to a file on disk and writing to a socket causes different amount of load apparently. For a socket it also depends on the socket options. Given all of the above one of the things we could do is make ZipTricks output a test archive through a webserver and measure that for example, with an end-to-end benchmark that uses curl or something like that. That would take more time. We could also say that, for Rack output at least, we would apply a blanket roughly-socket-sized buffer, and for other cases the user of the library has to apply their own buffering if they see the need to do so - and delete all the "intermediate" buffering changes I have introduced.

The example for S3 writes will work if we do not apply optimisations with bypassing writes, but I can also remove that example altogether because it does not contribute to solving the case that initiated this work (#78) it is more something of a nice thing that once popped up in the Github issues for the library (see #45). So I guess these are the questions that remain:

* How do we optimise for the right thing? In this case "the right thing" seems to be specifically usage in a webserver context 🤔
* Is it really that good of a plan to let zlib maintain a buffer of 5MB per ZIP? If you have a webserver which outputs to multiple clients at the same time it does add up. We use a 5MB buffer in a different way on our download pipe and it does mean that if we want to service 300 people we need to reserve 1.5G of memory just for chunks in flight. Same applies to zip_tricks deflate buffer. To be honest I would set it smaller, even if the compression would turn out less efficient due to less table reuse.

What are your thoughts? I don't know if we can spin up an archiving puma and `ab` against it in a CI context, but we could do it in the `bench/` directory and try it that way...

felixbuenemann:
We only need a single small write buffer, lets say 16K or 64K in front of the final IO/Sink.

I already have an optimized version of the `WriteBuffer` with a double sized buffer that is 5x faster than master and does not reallocate memory. In most cases it guarantees that the chunks it writes are at least as big as the buffer case, which would be fine for the S3 case.

**BUT:** The buffer has a bypass case, if the incoming write is larger than the buffer size, it flushes the buffer and passes the write though to the underlying IO.

This is the case were it cannot guarantee the minimum chunk size and splitting large chunks is expensive, so I need to find a good way to handle that. One idea is to only apply the fast-path if the buffer is currently empty and otherwise byte the bullet and split the chunk hoping the following writes will be big as well. The other solution is to exceed the buffer capacity, but that causes unbounded growth of the buffer memory, since we can't control how much bigger than the buffer size it is.

The real problem here is the passthrough behavior of the `StoredWriter`, since users can do unbounded writes. The `DeflatedWriter` will never write more than 64K with default mem level and 128K with max mem level (not currently used), so that is easy to work with.

As soon as we chunk those up internally, we end up allocating large amounts of new strings and sadly slices of strings are immediately allocated and not copy-on-write like when duping a complete string.

felixbuenemann:
Oh, forgot to post the optimized version of the `WriteBuffer`:

```ruby
class WriteBuffer
  def initialize(writable, buffer_size)
    @buf = ("\0".b * (buffer_size * 2)).clear
    @buffer_size = buffer_size
    @writable = writable
  end

  def <<(data)
    len = data.bytesize
    cap = @buffer_size
    if len >= cap
      flush unless @buf.empty? # <- this is were we can output less than @buffer_size
      @writable << data
    else
      @buf << data
      flush if @buf.bytesize >= cap
    end
    self
  end

  def flush
    unless @buf.empty?
      @writable << @buf
      @buf.clear
    end
    self
  end
end
```

julik:
I've put this together as an attempt to see if we can measure what actually matters:

https://gist.github.com/julik/bfc893fa8e29d646b3656185634fe1a8

It shows:

```
Calculating -------------------------------------
Rack each with bufsize of 0
                          0.040  (± 0.0%) i/s -      1.000  in  25.070461s
Rack each with bufsize of 1024
                          0.041  (± 0.0%) i/s -      1.000  in  24.373276s
Rack each with bufsize of 4096
                          0.049  (± 0.0%) i/s -      1.000  in  20.447174s
Rack each with bufsize of 16384
                          0.051  (± 0.0%) i/s -      1.000  in  19.788060s
Rack each with bufsize of 65536
                          0.050  (± 0.0%) i/s -      1.000  in  19.881594s

Comparison:
Rack each with bufsize of 16384:        0.1 i/s
Rack each with bufsize of 65536:        0.1 i/s - 1.00x  slower
Rack each with bufsize of 4096:        0.0 i/s - 1.03x  slower
Rack each with bufsize of 1024:        0.0 i/s - 1.23x  slower
Rack each with bufsize of 0:        0.0 i/s - 1.27x  slower
```

It is clear from it that some amount of buffering really does help with a webserver context, as the original issue stipulated. I think I am fine with removing the "flush no less than requested buffer size" limitation and removing the S3 example. This will make the job easier, let us achieve our objectives and also avoid too many allocations and slices. We probably should also mention in the code comments that we are not doing an `Array` of strings given to us to be later `join`-ed because we want to allow retained strings to be given over and over.

Will that work?

julik:
Let's implement your version of the buffer, remove buffering embedded in `Streamer` and keep the buffer in the `OutputEnumerator`.

felixbuenemann:
That sounds like a good plan. Sorry that I didn't get around to it yet as promised, but my Apple M1 MacBook Air arrived a week early on Friday and I couldn't help hacking around with it.

I've already got a native arm64 Ruby 2.7 environment, but JRuby and MRI 2.1 are still WIP. I've been busy with upstream patches, fixing OpenSSL 1.1.1 arm64 support etc.

felixbuenemann:
Oh and it is a fun device and has crazy performance, but for development very early adopter stage.

felixbuenemann:
@julik If you're impatient, just take my buffer code and roll with it.

julik:
Alright

spec/zip_tricks/write_buffer_spec.rb
```diff
@@ -10,28 +10,83 @@
...
 
+    adapter.flush
+  end
+
+  it 'bypasses larger writes, even if the amount of data accumulated is smaller than bufsize' do
+    # The WriteBuffer reuses strings, so examining its output is easier via Arrays
+    # if duplication gets applied on every write
+    class Duplicator < Struct.new(:accumulator)
+      def <<(data)
+        accumulator << data.dup; self
```
felixbuenemann, line 29: Rubocop doesn't like this:

```
spec/zip_tricks/write_buffer_spec.rb:29:32: C: Style/Semicolon: Do not use semicolons to terminate expressions.
586        accumulator << data.dup; self
```

julik: This should have been fixed in the squash, let me double-check

julik: Yep that was still on master, thanks for catching. Fixed it there

PR #85, opened 2020-11-10
Fix Specs for JRuby / TruffleRuby

felixbuenemann:
* Fix missing `RangeError` rescue in `ZipTricks::FileError` (actual bugfix, also applies to MRI when offset >= 1**63)
* Skip allocation tracing checks on JRuby (not supported) and TruffleRuby (crashes)
* The block deflate spec did excessive string allocations due to operator precedence (fixed with brackets)
* The remote uncap spec used spawn, which doesn't work with JRuby – replaced with `IO.popen`

The following are improvements unrelated ro JRuby/TruffleRUby compatibility:
* RemoteUncap: Remove hardcoded timeout, wait for server to be ready by reading stdout (as used by puma's own integration tests)
* RemoteUncap: Remove hardcoded TCP port that could conflict, use a free one from ephemeral port range (we can't just grep for "Listening on http://127.0.0.1:5432", since until Puma 4.x the proper port wasn't shown and ruby 2.1 has to use 3.11.x)

### DIFF
--- a/spec/zip_tricks/remote_uncap_spec.rb
+++ b/spec/zip_tricks/remote_uncap_spec.rb
@@ -4,7 +4,13 @@ require 'net/http'
 describe ZipTricks::RemoteUncap do
   before :all do
     rack_app = File.expand_path(__dir__ + '/remote_uncap_rack_app.ru')
-    command = 'bundle exec puma --port 9393 %s' % rack_app
+    # find a free tcp port
+    tcpserver = TCPServer.new('127.0.0.1', 0)
+    port = tcpserver.addr[1]
+    addr = tcpserver.addr[3]
+    tcpserver.close
+    @server_addr = "#{addr}:#{port}"
+    command = %W[bundle exec puma --bind tcp://#{@server_addr} #{rack_app}]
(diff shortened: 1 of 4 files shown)


spec/zip_tricks/remote_uncap_spec.rb
```diff
@@ -4,21 +4,34 @@
...
+      Process.wait(@server_pid)
+    rescue Errno::ECHILD
+    end
   end
 
   after :each do
     begin
       File.unlink('temp.zip')
-    rescue
-      Errno::ENOENT
```
felixbuenemann, line 21: This was rescuing `StandardError` and returned the constant `Errno::ENOENT`. Surely not the intended behavior.

julik (approved):
😍

felixbuenemann:
The long queue times on Travis CI are killing me, I'm working on GitHub Actions setup right now.

felixbuenemann:
GitHub Actions is working an running in under 2mins with near zero queue times. According to those tests I need to bump JRuby from 9.0 to 9.2 here, since even 9.1 is crashing with null pointer exceptions.

PR #86, opened 2020-11-10
Replace macOS Archive Utility tests with libarchive

felixbuenemann:
Since Archive Utility is a GUI app that is hard to automate and uses the systems' `libarchive2.dylib` under the hood, we just test against libarchive directly using the ffi-libarchive gem.

Since libarchive is available on Linux, we can run those specs in CI as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 rvm:
 - 2.1.0
 - 2.6.5
-- jruby-9.0
+- jruby-9.2
 sudo: false
+addons:
+  apt:
+    packages:
+      - libarchive-dev
 cache: bundler
 script:
   - bundle exec rake
--- a/spec/support/zip_inspection.rb
+++ b/spec/support/zip_inspection.rb
@@ -10,20 +10,4 @@ module ZipInspection
     escaped_cmd = Shellwords.join([zipinfo_path, '-tlhvz', path_to_zip])
     $zip_inspection_buf.puts `#{escaped_cmd}`
   end
-
-  def open_with_external_app(app_path, path_to_zip, skip_if_missing)
-    bin_exists = File.exist?(app_path)
-    skip "This system does not have #{File.basename(app_path)}" if skip_if_missing && !bin_exists
-    return unless bin_exists
-    `#{Shellwords.join([app_path, path_to_zip])}`
-  end
-
-  def open_zip_with_archive_utility(path_to_zip, skip_if_missing: false)
-    # ArchiveUtility sometimes puts the stuff it unarchives in ~/Downloads etc. so do
-    # not perform any checks on the files since we do not really know where they are on disk.
-    # Visual inspection should show whether the unarchiving is handled correctly.
-    au_path = '/System/Library/CoreServices/Applications/Archive Utility.app/' \
-              'Contents/MacOS/Archive Utility'
-    open_with_external_app(au_path, path_to_zip, skip_if_missing)
-  end
 end
--- a/spec/zip_tricks/streamer_spec.rb
+++ b/spec/zip_tricks/streamer_spec.rb
@@ -1,5 +1,6 @@
 require_relative '../spec_helper'
 require 'complexity_assert'
+require 'ffi-libarchive'
 
 describe ZipTricks::Streamer do
   let(:test_text_file_path) { File.join(__dir__, 'war-and-peace.txt') }
@@ -182,7 +183,7 @@ describe ZipTricks::Streamer do
   end
 
   it 'creates an archive that OSX ArchiveUtility can handle' do
-    outbuf = Tempfile.new('zip')
+    outbuf = StringIO.new
     outbuf.binmode
 
     zip = ZipTricks::Streamer.new(outbuf)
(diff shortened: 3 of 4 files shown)


felixbuenemann:
Run `otool -L /System/Library/CoreServices/Applications/Archive\ Utility.app/Contents/MacOS/Archive\ Utility` to see that it is linked against `/usr/lib/libarchive2.dylib`, which is the same lib we're loading when using `ffi-libarchive`.

felixbuenemann:
@julik This is an alternative to #83. What do you think?

.travis.yml
```diff
@@ -1,11 +1,15 @@
 rvm:
 - 2.1.0
 - 2.6.5
-- jruby-9.0
+- jruby-9.2
 sudo: false
+addons:
+  apt:
+    packages:
+      - libarchive-dev
```
felixbuenemann, line 9: I used `libarchive-dev` here instead of `libarchive13`, because on Ruby 2.1.0 we install an ancient 0.2.0 version of ffi-libarchive and it tries to load `libarchive.so.2` instead of `libarchive.so.13`, but does a fallback to `libarchive.so` which is installed by the dev package.

julik:
It doesn't seem like my `libarchive` is compatible (Mojave):

```julik@jet zip_tricks (replace-archive-utility-with-libarchive-tests) $ bspec

An error occurred while loading ./spec/zip_tricks/streamer_spec.rb. - Did you mean?
                    rspec ./spec/zip_tricks/stream_crc32_spec.rb
                    rspec ./spec/zip_tricks/zip_writer_spec.rb
                    rspec ./spec/zip_tricks/file_reader_spec.rb

Failure/Error: require 'ffi-libarchive'

FFI::NotFoundError:
  Function 'archive_read_open1' not found in [libarchive.dylib]
# ./spec/zip_tricks/streamer_spec.rb:3:in `require'
# ./spec/zip_tricks/streamer_spec.rb:3:in `<top (required)>'
Run options: exclude {:interactive=>true}


Finished in 0.00004 seconds (files took 0.48357 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples

julik@jet zip_tricks (replace-archive-utility-with-libarchive-tests) $ ```

**Comment by felixbuenemann** [comment]
@julik Mojave is quite ancient, can you try with `'ffi-libarchive', '0.2.0'` in the gemspec?

**Comment by julik** [comment]
So, on the positive side: yes that does help, tests pass. On the negative side: Archive Utility on Mojave does not use libarchive, it uses something else (I suspect either the Bom.framework or PackageKit.framework):

```
julik@jet zip_tricks (replace-archive-utility-with-libarchive-tests) $ otool -L /System/Library/CoreServices/Applications/Archive\ Utility.app/Contents/MacOS/Archive\ Utility
/System/Library/CoreServices/Applications/Archive Utility.app/Contents/MacOS/Archive Utility:
	/System/Library/PrivateFrameworks/PackageKit.framework/Versions/A/PackageKit (compatibility version 1.0.0, current version 434.0.0)
	/System/Library/Frameworks/PreferencePanes.framework/Versions/A/PreferencePanes (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 23.0.0)
	/System/Library/PrivateFrameworks/Bom.framework/Versions/A/Bom (compatibility version 2.0.0, current version 195.0.0)
	/System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 963.250.1)
	/System/Library/PrivateFrameworks/DiskImages.framework/Versions/A/DiskImages (compatibility version 1.0.8, current version 480.255.1)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1575.10.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.4)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
	/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit (compatibility version 45.0.0, current version 1671.50.101)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1575.10.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 944.3.0)
```

There is also some interesting behaviour we had people report in the wild. For example, on Mojave at least, if your ZIP contains a .docx file in stored mode (which is itself a ZIP), the ArchiveUtility will try to unarchive _that_ .docx as part of your archive. And won't unarchive the "outer" items correctly. Presumably because it does "straight ahead" reading instead of scanning the central directory first (like it should). This might very well be different on newer OS versions as they use functions from libarchive - as you noted.

Given all of that, I actually wonder: is this testing that valuable for ZipTricks after all these years? We already wiped out all the `qa/` stuff and we do have Rubyzip as an "external" verification mechanism. It might so happen that for peculiar things like blockwise deflate that we have we might want to test with ArchiveUtility, but is it necessary for every test run? 🤔

felixbuenemann:
I would be in favor of dropping automated Archive Utility tests.

julik:
Let’s do it then.

PR #89, opened 2020-11-11
GitHub Actions CI

felixbuenemann:
This contains the workflow configuration to run the CI on GitHub Actions. 

Results are usually ready in 1-2 minutes. 

Software versions have been slightly bumped compared to the current Travis CI setup:

* Ubuntu 18.04 Bionic
* Ruby 2.7
* Ruby 2.1
* JRuby 9.2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   lint:
     name: Code Style
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


felixbuenemann:
Made the workflow a bit smarter so it doesn't run two builds, if both push and pull_request events are triggered.

felixbuenemann:
I'll do another force push with no changes, jsut to show speedup through bundler caching and skipping.

felixbuenemann:
So with warm bundler caches:

* Rubocop: 15s
* Ruby 2.7: 30s
* Ruby 2.1: 45s
* JRuby: 60s

The "skip bundler if Gemfile.lock hasn't changed" is most effective von JRuby.

So feedback cycle is much improved, also since each check is immediately visible on the PR.

julik:
@felixbuenemann thank you! Looks good, but I really need to verify this internally since we are company-standardized on Travis (have been for some time). We might also have the option of moving this under our .com Travis account but I don't know if this is going to preserve contributor access to the CI results.

Since at the moment I am on holiday - can we hold this matter until next week Wednesday?

felixbuenemann:
Sure, no problem. 

Part of the reason for the Travis CI slowdowns seems to be ongoing problems with Google Compute Engine and there are thousands of jobs in the backlog. See https://www.traviscistatus.com

felixbuenemann:
@julik Maybe we can enable this in addition to Travis CI? At least until they figure out their problems.

julik:
@felixbuenemann We have a go on this - remove the Travis YAML please and I'll merge. We should be mindful that we have a company credit on GH actions which ZipTricks will now contribute to. If we find it uses too much resources we'll need to revert to Travis.

felixbuenemann:
Great, I'm using a maximum of caching, so impact on budget should be very low.

Since zip_tricks ia a public open source repo, shouldn't it be excempt form the company budget?

felixbuenemann:
@julik I've removed the `.travis.yml` and rebased, this is ready to merge.

Btw. the test suite uses over 1 GB RAM, mostly because there are huge `StringIO` objects with 50 MB and larger buffers. This caused the Travis CI workers to sometimes be killed due to hitting the memory limit.

I have a WIP branch that brings down memory usage and further speeds up the specs, because those cause heavy GC work and slow down compression. I haven't worked on it in a week or so, but it was pretty far along. Mostly what I did was to scale down test strings and sometimes I switched to `ManagedTempfile`.

I haven't fully completed it, but I think somewhere in the 300 MiB range should be feasible.

julik (approved):
🚀  Let's see how this works out credits-wise.

In terms of memory use testing, what would be our options? We can write to Tempfiles in tests and verify from there, we can also add some tests for the number of String allocations, but in a managed language we could only do so much. Curious to see your PoC.

I also have 32G so I wouldn't have noticed the test suite exploding the memory use.

felixbuenemann:
@julik See #98 for the work on removing memory usage and reducing runtime for the specs.

PR #95, opened 2020-11-17
Don't change source encoding in WriteAndTell and BlockWrite

felixbuenemann:
Since `WriteAndTell#binary_bytes` and `BlockWrite#<<` were calling `String#force_encoding` on the written string, they were changing the encoding of the source strings.

This fixes the bugs and specs that accidentally depended on the mutation.

### DIFF
--- a/lib/zip_tricks/block_write.rb
+++ b/lib/zip_tricks/block_write.rb
@@ -23,20 +23,7 @@ class ZipTricks::BlockWrite
     # Zero-size output has a special meaning  when using chunked encoding
     return if buf.nil? || buf.bytesize.zero?
 
-    # Ensure we ALWAYS write in binary encoding.
-    encoded =
-      if buf.encoding != Encoding::BINARY
-        # If we got a frozen string we can't force_encoding on it
-        begin
-          buf.force_encoding(Encoding::BINARY)
-        rescue
-          buf.dup.force_encoding(Encoding::BINARY)
-        end
-      else
-        buf
-      end
-
-    @block.call(encoded)
+    @block.call(buf.b)
     self
   end
 end
--- a/lib/zip_tricks/write_and_tell.rb
+++ b/lib/zip_tricks/write_and_tell.rb
@@ -10,9 +10,8 @@ class ZipTricks::WriteAndTell
 
   def <<(bytes)
     return self if bytes.nil?
-    binary_bytes = binary(bytes)
-    @io << binary_bytes
-    @pos += binary_bytes.bytesize
+    @io << bytes.b
+    @pos += bytes.bytesize
     self
   end
 
@@ -22,14 +21,5 @@ class ZipTricks::WriteAndTell
 
   def tell
     @pos
-  end
-
-  private
-
-  def binary(str)
-    return str if str.encoding == Encoding::BINARY
-    str.force_encoding(Encoding::BINARY)
-  rescue RuntimeError # the string is frozen
-    str.dup.force_encoding(Encoding::BINARY)
   end
 end
--- a/spec/zip_tricks/block_write_spec.rb
+++ b/spec/zip_tricks/block_write_spec.rb
@@ -44,6 +44,15 @@ describe ZipTricks::BlockWrite do
     blobs.each { |s| expect(s.encoding).to eq(Encoding::BINARY) }
   end
 
+  it 'does not change the encoding of source strings' do
+    hello = 'hello'.encode(Encoding::UTF_8)
+    accum_string = ''.force_encoding(Encoding::BINARY)
+    adapter = described_class.new { |s| accum_string << s }
+    adapter << hello
+    expect(accum_string.encoding).to eq(Encoding::BINARY)
+    expect(hello.encoding).to eq(Encoding::UTF_8)
+  end
+
   it 'omits strings of zero length' do
     blobs = []
     adapter = described_class.new { |s| blobs << s }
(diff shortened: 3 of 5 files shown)


julik:
While I understand the idea, I doubt it is a good plan to force people to feed ZipTricks strings in binary encoding and no other. It can easily put them on a  bug hunt because there is no guarantee that the string that you want to append to a file comes from you. Imagine you are in UTF-8 and you are outputting a CSV through a CSV writer. Do you have a guarantee that it will correctly tag the strings it outputs with the right encoding?

felixbuenemann:
I didn't change any behavior, it still doers change the encoding to binary, but without changing the source.

felixbuenemann:
@julik Look closely, I'm calling `String#b`.

felixbuenemann:
@julik The reason I changes the specs to use `File.binread` was because they were using binary offsets via `String#slice`, that only worked because the source string was mutated. The files have an UTF-8 BOM, so even `String#byteslice(0...5)` would retun a wrong result, since it would skip the URF-8 BOM and use byte offsets from there, but the `StringIO` will noever skip the BOM. So the only way to make the tests have the same offsets is to use binary encoding on the source string.

felixbuenemann:
Sorry, `String#byteslice` doesn't skip the BOM, you just don't see it, since it is a non-printable character:

```ruby
x = "\xEF\xBB\xBFThe Proj"
x.slice(0...5)
=> "The "
x.slice(0...5).b
=> "\xEF\xBB\xBFThe "
x.byteslice(0...5)
=> "Th"
x.byteslice(0...5).b
=> "\xEF\xBB\xBFTh"
s = StringIO.new(x)
s.read(5)
=> "\xEF\xBB\xBFTh"
```

So anyways, if you want to compare `StringIO#read` and `String#slice` results with the same offsets, they must both use the same encoding, bei it `UTF-8` or `ASCII-8BIT`.

julik:
Ok got it. I think we should add a test that ensures the strings we send through to our writers get conformed to binary, and then we can deal with the other aspects. What I am worried about is people starting to get encoding errors just when writing content into their ZIPs - many don't know how to handle those errors and they are a nuisance, since ZT is at a boundary I think we ought to be dealing with this. Maybe add a wrapper which does `#b` on every String passing through and at least decouple it from the others?

felixbuenemann:
I think the right place to do it is where the binary data is actually needed.

There is no obvious reason for `WriteAndTell` to do a conversion to binary and it currently does it multiple times, because both the `StoredWriter` and the `Streamer` wrap the IO in it.

Anything going through Zlib.deflate doesn't care about encoding at all, since it goes to C-land that is blissfully unaware of encodings and comes back out as binary.

This leaves the `StoredWriter` as the only place where it should actually convert to binary.

If you take this code out of `WriteAndTell` and notice that `advance_position_by` is only needed by the `Streamer`, the class is left with basically `@pos += data.bytesize` which you can just inline and delete the class alltogether…

felixbuenemann:
To keep it short, I don't think there is anything gained by adding yet another wrapper class doing:

```ruby
class ZipTricks::BinaryEncoder
  def initialize(io)
    @io = io
  end

  def <<(data)
    @io << data.to_s.b
  end
end
```

All those wrapper classes add both cognitive and performance overhead and IMHO should be avoided for trivial code.

julik:
Ok let me play with this branch for a moment

felixbuenemann:
No prob, I'm pushing a spec for the mutation test. Then I'm through.

felixbuenemann:
@julik OK, added the specs, go ahead with your changes.

We can also jsut merge this and do those in a separate PR, since they will likey spiral of from the bugfix nature of this PR.

julik (approved):
Yeah I've played a bit with the branch and this seems to work too:

```ruby
  it 'accepts data in any encoding and outputs it to a binary encoded IO without raising errors' do
    encodings = [Encoding::BINARY, Encoding::UTF_8, Encoding::UTF_16, Encoding::ISO8859_1]
    str = "Mary had a little lamb and a little 🎠"
    output = StringIO.new
    output.set_encoding(Encoding::BINARY)
    expect {
      ZipTricks::Streamer.open(output) do |zip|
        encodings.each do |encoding|
          zip.write_stored_file("text-stored #{encoding}.bin") do |sink|
            sink << str.force_encoding(encoding)
            sink << str.encode(encoding, undefined: :replace, invalid: :replace)
          end
          zip.write_deflated_file("text-compressed #{encoding}.bin") do |sink|
            sink << str.force_encoding(encoding)
            sink << str.encode(encoding, undefined: :replace, invalid: :replace)
          end
        end
      end
    }.not_to raise_error
  end
```

I'll merge, just let's keep an eye out in case someone complains they started getting encoding errors, as this is a sensitive area

felixbuenemann:
> I'll merge, just let's keep an eye out in case someone complains they started getting encoding errors, as this is a sensitive area

Not sure what you mean with that comment.

I'm still converting everything to binary encoding.

I could've also written `@io << data.dup.force_encoding(Encoding::BINARY)`, but `@io << data.b` is shorter and does it in one method call. Not sure if my previous comment din't make that clear.

I've removed the special cases for frozen and binary encoded strings because a `.dup` is basically free due to copy-on-write discussed earlier.

There will be some bytes allocated to hold the new string object referencing the existing data, but we get rid of the branching, so that's probably a win.

julik:
> Not sure what you mean with that comment.

What I mean is that we should watch closely for issues mentioning encoding errors when we change the treatment of the strings :-)

PR #96, opened 2020-11-17
Optimize DeflatedWriter

felixbuenemann:
* Drop interim flushing that will never trigger (Zlib will flush at roughly 4M with the default `Zlib::DEF_MEM_LEVEL` and 8M with `Zlib::MAX_MEM_LEVEL`, so flushing on 5M boundaries is useless)
* Use `Zlib::Streamer` methods to get compressed and uncompressed size
* Use block form of `deflate` to lower memory usage (especially when writing with `IO#each(nil, block_size) { |chunk| zip << chunk }`)
* Fix missing `Zlib::Deflate#close` call that caused about 250K mem leak per instance


julik:
Excellent! And I will backport this to 4.x for the #close fix

julik (approved):

PR #97, opened 2020-11-24
Remove complexity_assert development dependency

felixbuenemann:
This was used by the streamer spec to assert linear complexity, but causes a huge amount of array allocations with 300 MB memory usage and takes up about 40% of the entire test suite runtime for a single spec.

Instead replace it with a simple inline implementation that asserts linear complexity with 75% confidence interval.

While local tests show that the usual variance is within 5% or 95% confidence, the lower confidence is needed to account for fluctuation between test runs caused by GC or system load.

This could be counted by going for higher file counts, but starting from about 10K files that makes the spec unbearably slow.

### DIFF
--- a/spec/zip_tricks/streamer_spec.rb
+++ b/spec/zip_tricks/streamer_spec.rb
@@ -1,5 +1,4 @@
 require_relative '../spec_helper'
-require 'complexity_assert'
 
 describe ZipTricks::Streamer do
   let(:test_text_file_path) { File.join(__dir__, 'war-and-peace.txt') }
@@ -29,21 +28,35 @@ describe ZipTricks::Streamer do
   end
 
   it 'has linear performance depending on the file count' do
-    module FilecountComplexity
-      def self.generate_args(size)
-        [size]
-      end
-
-      def self.run(n_files)
+    scale = 2 # scaling factor between tests
+    tries = 4 # retries to reduce variance
+    max_variance = 0.15 # 15% fuzz factor
+    # 4, 8, 16, ..., 4096
+    results = (0..11).map do |n|
+      n_files = scale**n
+      tries.times.reduce(scale.to_f) do |last|
+        start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
         ZipTricks::Streamer.open(ZipTricks::NullWriter, writer: FakeZipWriter.new) do |w|
           n_files.times do |i|
             w.write_stored_file(format('file_%d', i)) { |body| body << 'w' }
           end
         end
+        result = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+        variance = (last - result).abs / result
+        break result if variance <= max_variance
+        result
       end
     end
 
-    expect(FilecountComplexity).to be_linear
+    # Calculate the factor compared to the previous result.
+    (results.count - 1).downto(1).each do |i|
+      results[i] = results[i] / results[i - 1]
+    end
+
+    median = results.sort[results.count / 2]
+    confidence = 0.25 # 75% confidence interval
+
+    expect(median).to be_within(scale * confidence).of(scale)
   end
 
   it 'raises an InvalidOutput if the given object does not support the methods' do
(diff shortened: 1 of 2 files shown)


felixbuenemann:
Argh, I rearranged the blocks and am measuring file size instead of execution time, will fix.

felixbuenemann:
@julik Another approach would be to tag the slow spec and only run it on demand.

The code for this ended up much more complex that my initial PoC to avoid flipping due to variance.

felixbuenemann:
### Performance comparison

RAM: 468M -> 33M (14x reduction)
CPU: 7.19s -> 0.56s (13x reduction)

**Before:**

```
$ /usr/bin/time -l rspec spec/zip_tricks/streamer_spec.rb:31
Run options: include {:locations=>{"./spec/zip_tricks/streamer_spec.rb"=>[31]}}
.

Finished in 6.76 seconds (files took 0.13213 seconds to load)
1 example, 0 failures

        7.19 real         6.84 user         0.32 sys
           491335680  maximum resident set size
```

**After:**

```
$ /usr/bin/time -l rspec spec/zip_tricks/streamer_spec.rb:31
Run options: include {:locations=>{"./spec/zip_tricks/streamer_spec.rb"=>[31]}}
.

Finished in 0.19886 seconds (files took 0.11675 seconds to load)
1 example, 0 failures

        0.56 real         0.46 user         0.08 sys
            34746368  maximum resident set size
```

felixbuenemann:
I have already seen a case were the 25% variance check was not enough on JRuby, although by a tiny margin. Something like rspec-retry could be a solution, though it is more of a bandaid.

felixbuenemann:
@julik Personals I tend towards putting such slow and potentially flipping tests under a tag and only run it on demand.

This could also be done as a benchmark with benchmark-ips, which implements quite a bit more logic to weed out variance across different test sizes than what I can do.

Any attempts to make them more reliable end up spending even more cpu time, since it comes down to weeding out variance through more tries or larger iteration counts. – I wen't down the rabbit hole with this one and you quickly spend seconds on a single test, where the hole remaining test suite can run in 2.5s without this spec.

There is also the case here that the behaviour is more like "eventually linear", if you look at the distribution for the tested range it goes from something like 50% to 150% of the 2x linear target. 

What do you think? – It would also be interesting to known what the original intention for adding the test was.

julik:
The original intention behind the test can be found in #14 - we had a situation where we inadvertently added a feature which would scan through the file list. As the file list would grow longer, adding a new file would scan the entire list again, which ultimately resulted in an "accidentally quadratic" situation. This started being very noticeable with even a few thousand files, and when this was fixed we decided that it is important to "codify" the fact that it is fixed (and should stay fixed) with a test.

Now, initially you will see that it also used a custom assertion performance-wise, but we went with complexity-assert for the following reason: for someone just getting into the code of the library, it would be conceptually easier to understand what the code does, as it calls into a well-defined, documented dependency. I do understand that there is value in having faster tests, and we could make this test skip under most CI runs and only keep it in, say, one of the CI passes.

But I don't know whether it is very valuable to brutally optimise the performance of tests, knowing that we will have to sacrifice some amount of readability/clarity while doing so. In the library itself such optimisations really are justified because the performance cost is paid by all users of the library (sometimes unknowingly), but in tests the priorities are not so clear-cut.

Can we try an approach where we isolate these "heavy-duty" tests under a single CI pass and only pay the runtime cost there? I think we should also be able to tweak the amount of items/iterations that complexity assert does.

@piotrmurach do you have suggestions maybe?

piotrmurach:
Thanks for tagging me, even if by accident. Enjoyed reading the discussion so far. Agree with @julik that keeping a perf test to capture potential regressions is useful. Otherwise, how will you ensure that some refactoring is not going to cause a regression? I like the idea of grouping 'heavy-duty' tests together. I tend to keep them in a separate folder and run with `spec:perf`. Separate job or workflow to run these is a good way to resolve this, on one hand, having a snappy base test suite and a backup of perf tests as required to merge changes. 

If you want to use 'out-of-the-box' matchers then [rspec-benchamark complexity matcher](https://github.com/piotrmurach/rspec-benchmark#14-complexity) would be a good candidate. It allows you to pick the sampling rate and threshold. If that feels too 'heavy' then you can try writing your own `be_linear` matcher based on the [benchmark-trend](https://github.com/piotrmurach/benchmark-trend). Would be curious to know how it compares with your current way.

felixbuenemann:
I will need to look into GitHub Actions documentation what would be good ways to trigger this. There are also scheduled test runs that could be used for this or a filter that for example only triggers on merge to master.

julik:
Sorry, I have renamed `master` so when you get back to this branch just rebase off `main` 🙏

PR #98, opened 2020-11-25
WIP: Speed Up Specs & Reduce Memory Usage

felixbuenemann:
This branchs is not fully ready and still needs some cleanup and there are a few specs left that allocate a few megs.


felixbuenemann:
**First Benchmarks:**

CPU: 15s to 4s (4x speedup)
RAM: 1.2G to 100M (13x speedup)

Though these contain the complexity assert removal, that gives a big speedup on its own.

julik:
Sorry, I have renamed `master` so when you get back to this branch just rebase off `main` 🙏

PR #99, opened 2020-11-25
Add Felix to the authors list

julik:
With this amount of contributions we should appoint him as one of the authors


felixbuenemann (approved):
LGTM

Issue #104, opened 2021-01-05
Ruby 3 compatibility


swiknaba:
Currently, specs are failing under Ruby 3, see: https://github.com/WeTransfer/zip_tricks/pull/103

We should address these issues, i.e. update the syntax to be compatible with Ruby 3.

julik:
I have merged #103 thank you for doing this work! We do not have ruby 3 in production anywhere so it is very valuable. Can you pick this up once `rspec-mocks` is updated?

swiknaba:
rspec-mocks is updated now (details: https://github.com/rspec/rspec-mocks/issues/1396), basically we'll need to use `>= 3.10.2` for Ruby 3 compatibility.

PR #105, opened 2021-02-16
upgrade rspec-mock (Ruby 3 compatibility, we can now disallow failures for Ruby 3 on CI)

swiknaba:
Had to hard code the minimum `rspec-mock` gem version (dev dependency). Using `>= 3.10.2` all specs with Ruby 3.0 pass.

Possibly we could use [appraisal](https://github.com/thoughtbot/appraisal/) to run different requirements for different Rubies for the CI. That'll require quite some config though.

closes https://github.com/WeTransfer/zip_tricks/issues/104

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,13 +58,11 @@ jobs:
     strategy:
       matrix:
         ruby:
+          - '3.0'
           - '2.7'
           - '2.1'
           - 'jruby-9.2'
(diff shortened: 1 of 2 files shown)


julik:
There is a way to do this without using appraisal - we would need to set up 2 builds which use different Gemfiles, I've used this setup in the past. Can you add them in a `gemfiles` directory under this PR and I'll try to configure?

PR #106, opened 2021-06-03
Allow UNIX permissions for entries to be configured

julik:
Since we now have permissions in Storm it makes sense to support them natively. This will also allow the fork in https://github.com/retrospectinc/zip_tricks to be dropped, since this is a feature they had to add

### DIFF
--- a/lib/zip_tricks/zip_writer.rb
+++ b/lib/zip_tricks/zip_writer.rb
@@ -137,7 +137,7 @@ class ZipTricks::ZipWriter
                                           crc32:,
                                           filename:,
                                           unix_permissions: nil
-                                          )
+                                         )
     # At this point if the header begins somewhere beyound 0xFFFFFFFF we _have_ to record the offset
     # of the local file header as a zip64 extra field, so we give up, give in, you loose, love will always win...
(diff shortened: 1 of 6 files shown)


lib/zip_tricks/zip_writer.rb
```diff
@@ -200,11 +192,15 @@ def write_central_directory_file_header(io:,
 
     # Because the add_empty_directory method will create a directory with a trailing "/",
     # this check can be used to assign proper permissions to the created directory.
-    io << if filename.end_with?('/')
-      [EMPTY_DIRECTORY_EXTERNAL_ATTRS].pack(C_UINT4)
+    # external file attributes        4 bytes
+    external_attrs = if filename.end_with?('/')
```
lorenzograndi4, line 196: Is that the only way to differentiate a directory from a file? What if a file's filename just ends with `/` ?

lorenzograndi4 (approved):

lib/zip_tricks/zip_writer.rb
julik: In zip_tricks it is the only way yes. File paths cannot end with a `/`

PR #111, opened 2022-02-03
Add custom comments to Streamer

pyrareae:
This is a tiny patch to allow setting the commit message.

### DIFF
--- a/lib/zip_tricks/zip_writer.rb
+++ b/lib/zip_tricks/zip_writer.rb
@@ -246,7 +246,9 @@ class ZipTricks::ZipWriter
   # @param num_files_in_archive[Fixnum] How many files the archive contains
   # @param comment[String] the comment for the archive (defaults to ZIP_TRICKS_COMMENT)
   # @return [void]
-  def write_end_of_central_directory(io:, start_of_central_directory_location:, central_directory_size:, num_files_in_archive:, comment: ZIP_TRICKS_COMMENT)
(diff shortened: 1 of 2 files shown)


lib/zip_tricks/zip_writer.rb
```diff
@@ -246,7 +246,9 @@ def write_data_descriptor(io:, compressed_size:, uncompressed_size:, crc32:)
   # @param num_files_in_archive[Fixnum] How many files the archive contains
   # @param comment[String] the comment for the archive (defaults to ZIP_TRICKS_COMMENT)
   # @return [void]
-  def write_end_of_central_directory(io:, start_of_central_directory_location:, central_directory_size:, num_files_in_archive:, comment: ZIP_TRICKS_COMMENT)
+  def write_end_of_central_directory(io:, start_of_central_directory_location:, central_directory_size:, num_files_in_archive:, comment: nil)
+    comment ||= ZIP_TRICKS_COMMENT
```
grdw, line 250: I feel the old code actually already achieved what you had in mind. There's no need for memoizing this variable.

grdw (changes requested):
Thanks for your contribution. Please add a spec for this change.

pyrareae:
Sorry for the long delay I have had a *lot* going on. I still intend to finish this pr by adding tests and such though.

lib/zip_tricks/zip_writer.rb
pyrareae: As I recall this was needed to handle a weird case with the default not being handled correctly.

